PR #13, opened 2021-11-03
Updating messaging on README

MichaelJolley:


README.md
```diff
@@ -5,7 +5,7 @@
 Official Node.js SDK for [Deepgram](https://www.deepgram.com/)'s automated
 speech recognition APIs.
 
-> This SDK only supports the new beta Deepgram API.
+> This SDK only supports hosted usage of api.deepgram.com.
```
briancbarrow, line 8: What does "hosted usage" mean? Do we need to explain that more?

briancbarrow (approved):

README.md
MichaelJolley: It means "not on-prem". I just copy/pasted a sentence Natalie asked to use instead of what was there.

PR #19, opened 2021-12-16
Added the ability to create temporary keys

MichaelJolley:
Updated the `keys.create` function to allow new `expirationDate` or `timeToLive` values. These are optional and one at most can be provided. Providing both will throw an error.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Updated
+
+- Updated the `keys.create` function to allow new `expirationDate` or `timeToLive`
+values. These are optional and one at most can be provided. Providing both will
+throw an error.
+
 ---
 
 ## [1.1.0]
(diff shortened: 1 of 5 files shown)


phazonoverload (changes requested):
Looks good to me - before approval can you show an example (even just in this thread) of creating a key with `expiration_date` and one with `time_to_live_in_seconds`?

MichaelJolley:
Sure thing.

Expiration Date
```js
deepgram.keys.create(PROJECT_ID, COMMENT, ["member"], new Date());
```


TTL
```js
deepgram.keys.create(PROJECT_ID, COMMENT, ["member"], null, 30);
```

phazonoverload:
Would it be feasible to allow that third param to be a number OR a date, and dependent on type it will act differently? Not sure if it's the right approach but welcome thoughts.

natalierutgers (approved):

MichaelJolley:
@phazonoverload I like that idea, but what would you name that parameter?

phazonoverload:
`expiry` I think works

MichaelJolley:
Updated based on our conversations to take an optional object that can only contain expirationDate or timeToLive

phazonoverload (approved):

phazonoverload:
Thanks so much for this @MichaelJolley

PR #24, opened 2022-02-09
fix: wss stream

igsova:
Modification on the url to work with stream for on-premise usage


MichaelJolley:
Thanks for the contribution @igsova. Unfortunately, this would break the SDK for all users of the new API. I'm also not sure it would work for on-prem users as the authentication method on the on-prem is different than the new API. That said, the SDK will work fine with on-prem deployments in the future when the on-prem images are updated to implement the new API.

PR #26, opened 2022-02-25
fix: fix types to match current api response

### LINKED ISSUE #25: Output of keys list doesn't match docs

https://developers.deepgram.com/sdks-tools/sdks/node-sdk/keys/#list-project-keys-example-request

These docs suggest this output when I list API keys for a project

```js
{
  api_keys: [
    {
      api_key_id: string,
      comment: string,
      created: string,
      scopes: Array<string>
    },
  ];
}
```

but i get this output

```js
{ 
  api_keys: [
    {
      member: {
        member_id: '...',
        email: '...
      },
      api_key: {
        api_key_id: '...',
        comment: '...',
        scopes: [...],
        created: '...'
      }
    },
    ...
  ]
}
```

lukeocodes:
updated the API key type to match the current API response

fixes #25

### DIFF
--- a/src/types/key.ts
+++ b/src/types/key.ts
@@ -1,43 +1,25 @@
 /**
  * API key used for authenticating with the Deepgram API
  */
-export declare type Key = {
+export type Key = {
   /**
-   * Api key member object
+   * Unique identifier of the key to use in API requests
    */
-  member: {
-    /**
-     * Unique identifier of the member on this API key
-     */
-    member_id: string;
-    /**
-     * Email of the member
-     */
-    email: string;
-  };
+  api_key_id: string;
   /**
-   * Api key object
+   * API key to send in API requests (Only displayed when first created)
    */
-  api_key: {
-    /**
-     * Unique identifier of the key to use in API requests
-     */
-    api_key_id: string;
-    /**
-     * API key to send in API requests (Only displayed when first created)
-     */
-    key?: string;
-    /**
-     * Comment for user reference
-     */
-    comment: string;
-    /**
-     * Timestamp of the date/time the key was created
-     */
-    created: string;
-    /**
-     * Array of scopes assigned to the key
-     */
-    scopes: Array<string>;
-  };
+  key?: string;
+  /**
+   * Comment for user reference
+   */
+  comment: string;
+  /**
+   * Timestamp of the date/time the key was created
+   */
+  created: string;
+  /**
+   * Array of scopes assigned to the key
+   */
+  scopes: Array<string>;
 };
--- a/src/types/key.ts
+++ b/src/types/key.ts
@@ -1,25 +1,43 @@
 /**
  * API key used for authenticating with the Deepgram API
  */
-export type Key = {
+export declare type Key = {
   /**
-   * Unique identifier of the key to use in API requests
+   * Api key member object
    */
-  api_key_id: string;
+  member: {
+    /**
+     * Unique identifier of the member on this API key
+     */
+    member_id: string;
+    /**
+     * Email of the member
+     */
+    email: string;
+  };
   /**
-   * API key to send in API requests (Only displayed when first created)
+   * Api key object
    */
-  key?: string;
-  /**
-   * Comment for user reference
-   */
-  comment: string;
-  /**
-   * Timestamp of the date/time the key was created
-   */
-  created: string;
-  /**
-   * Array of scopes assigned to the key
-   */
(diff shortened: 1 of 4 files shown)


briancbarrow (approved):

lukeocodes:
@MichaelJolley the question is if this should be a major version bump or whether we should parse the response and output the original format for compatibility.

What do you think?

src/types/key.ts
```diff
@@ -1,25 +1,43 @@
 /**
  * API key used for authenticating with the Deepgram API
  */
-export type Key = {
+export declare type Key = {
```
MichaelJolley, line 4: declare isn't needed here.

src/types/key.ts
```diff
@@ -1,25 +1,43 @@
...
   /**
-   * Unique identifier of the key to use in API requests
+   * Api key member object
```
MichaelJolley, line 6: API (capitalized since it's an abbreviation)

src/types/key.ts
```diff
@@ -1,25 +1,43 @@
...
    */
-  api_key_id: string;
+  member: {
```
MichaelJolley, line 8: The member object is nullable.

```ts
member?: {
```

src/types/key.ts
```diff
@@ -1,25 +1,43 @@
...
+     */
+    member_id: string;
+    /**
+     * Email of the member
+     */
+    email: string;
+  };
   /**
-   * API key to send in API requests (Only displayed when first created)
+   * Api key object
```
MichaelJolley, line 19: API (capitalized since it's an abbreviation)

MichaelJolley (changes requested):
I added comments, then realized this is the wrong way to fix this issue. In this PR, we're modifying the `Key` type. That type is currently used by the [KeyResponse](https://github.com/deepgram/node-sdk/blob/main/src/types/keyResponse.ts) type and the [Keys](https://github.com/deepgram/node-sdk/blob/main/src/keys.ts) static class as the return type of the create & get functions. Modifying the `Key` type would provide the incorrect return type for those methods. 

Instead, we should create a new `Member` type and add it as an optional property to the `KeyResponse` type, like:

```ts
import { Key } from "./key";
import { Member } from "./member";

/**
 * Response from the Deepgram API to list keys
 */
export type KeyResponse = {
  /**
   * Array of API keys associated with the project
   */
  api_keys: Array<{
    /**
    * Optional member associated with the API key
    */
    member?: Member;
    
    /**
    * API key
    */ 
    api_key: Key;
  }>;
};
```

MichaelJolley:
@lukeocodes answering the other question around SemVer: ideally we would retain the old typing and provide the previous output as well (if possible) and flag the properties we know are wrong/outdated as deprecated to be removed at a future date.

lukeocodes:
> I added comments, then realized this is the wrong way to fix this issue. In this PR, we're modifying the `Key` type. That type is currently used by the [KeyResponse](https://github.com/deepgram/node-sdk/blob/main/src/types/keyResponse.ts) type and the [Keys](https://github.com/deepgram/node-sdk/blob/main/src/keys.ts) static class as the return type of the create & get functions. Modifying the `Key` type would provide the incorrect return type for those methods. 
> 
> Instead, we should create a new `Member` type and add it as an optional property to the `KeyResponse` type, like:
> 
> ```ts
> import { Key } from "./key";
> import { Member } from "./member";
> 
> /**
>  * Response from the Deepgram API to list keys
>  */
> export type KeyResponse = {
>   /**
>    * Array of API keys associated with the project
>    */
>   api_keys: Array<{
>     /**
>     * Optional member associated with the API key
>     */
>     member?: Member;
>     
>     /**
>     * API key
>     */ 
>     api_key: Key;
>   }>;
> };
> ```

I'll look more closely at this. Something is yelling at me that this isn't the format being returned but I could be wrong.

Definitely considering tackling a test suit next

lukeocodes:
> @lukeocodes answering the other question around SemVer: ideally we would retain the old typing and provide the previous output as well (if possible) and flag the properties we know are wrong/outdated as deprecated to be removed at a future date.

I'm not sure I know how to achieve this without introducing a promise to wrap around _request in order to modify the output, with some property to switch to the new output?

lukeocodes:
Types for Key and Member now match the API response. I've changed the existing Member type as it didn't appear to be used anywhere else yet

MichaelJolley (approved):

PR #31, opened 2022-03-16
Feature/dev ed 30.add members endpoint

### LINKED ISSUE #30: Add Members Endpoint
## Proposed changes

We need to add members endpoint to SDK

https://developers.deepgram.com/api-reference/#members

briancbarrow:
closes #30 

I added the members endpoint to the SDK. 

I used similar method names as the `usage` endpoint was using. Let me know if I should change it to be just the verb (eg. `list` instead of `listMembers`)

I also updated the Members type object to having `first_name` and `last_name` instead of just `name` because that is what is coming back from the API. 

I added a generic Message type also for the responses that come back from DELETE calls. I noticed on the `keys` endpoint we are saying it returns a `void` but we actually return an object with a `message` property. I thought we might want to update that.

Do I need to write any tests surrounding this?

### DIFF
--- a/src/types/member.ts
+++ b/src/types/member.ts
@@ -1,6 +1,7 @@
 export type Member = {
   member_id: string;
-  name?: string;
+  first_name?: string;
+  last_name?: string;
   scopes?: Array<string>;
   email: string;
 };
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@deepgram/sdk",
-  "version": "1.2.1",
+  "version": "1.2.2",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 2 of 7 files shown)


src/types/member.ts
```diff
@@ -1,6 +1,7 @@
 export type Member = {
   member_id: string;
-  name?: string;
+  first_name?: string;
```
MichaelJolley, line 3: Because this type is also used by `key.ts` we can't modify it's structure. We'll likely need a new type for now.

MichaelJolley (changes requested):

MichaelJolley (changes requested):
Need to remove the member changes.

src/types/member.ts
briancbarrow: It looks like it is being used incorrectly though. When I call the api I get a `first_name` and `last_name` in the response. Not a `name` property. 

Screen Shot 2022-03-17 at 9 01 31 AM

briancbarrow:
changes will be merged in #37

PR #41, opened 2022-04-01
added browser live transcription option

briancbarrow:
- [x] Docs need to be up-to-date on DDP (the branch is created, just need to get the PR created and merged. Waiting on this to be approved in case there are any changes that need to be made still)
- [x] CHANGELOG.md needs to be updated in repo
- [x] package.json version needs to be updated to new version # (and run npm install to update package lock file)
- [ ] Tag the final commit with the version number
- [ ] Create a release in GH for that version number & changes


src/browser/index.ts
```diff
@@ -0,0 +1,32 @@
...
+   */
+  private _validateOptions() {
+    if (!this._apiKey || this._apiKey.trim().length === 0) {
+      throw new Error("DG: API key is required");
+    }
+
+    if (!this._apiUrl || this._apiUrl.trim().length === 0) {
+      throw new Error("DG: API url should be a valid url or not provided");
+    }
+  }
```
briancbarrow, line 31: Break this out into its own file and import it where it is used.

MichaelJolley:
@briancbarrow this looks really good. Couple things we need to address:

- [ ] Need to add browserify as a dev dependency. Without it, build fails.
- [ ] We need to figure out how to keep the type definitions. The process as is doesn't keep the types in the `dist` folder. So intellisense is lost for functions/types/etc.

Seriously good stuff here.

MichaelJolley (approved):

PR #46, opened 2022-04-27
Remove requirement for MIME type for local, pre-recorded audio

ambergeldar:


src/transcription/preRecordedTranscription.ts
```diff
@@ -44,15 +44,6 @@ export const preRecordedTranscription = async (
...
 
-  if (
-    !isUrlSource(source) &&
-    (source.mimetype === undefined || source.mimetype.length === 0)
-  ) {
-    throw new Error(
-      "DG: Mimetype must be provided if the source is a Buffer or a ReadStream"
-    );
-  }
-
```
briancbarrow, line 55: Do we need to remove the `Content-Type` header from line 61 below?

This is what we currently do:
```
const requestOptions: { [name: string]: { [key: string]: string } } = {};
  if (!isUrlSource(source)) {
    requestOptions.headers = {
      "Content-Type": source.mimetype,
    };
  }
  ```

briancbarrow (changes requested):
If we remove this requirement, should we remove the `Content-Type` header and also the source/mimetype property from the ReadableStreamSource and BufferSource types?

PR #48, opened 2022-06-14
Change ReadStreamSource to accept stream.Readable instead of only fs.ReadStream.

wdomburg:
So I know when I opened the original issue I requested that ReadStream be added as a TranscriptionSource, but on usage, I realize that it's useful to use the stream.Readable for more flexibility.

Since the code doesn't actually rely on added methods, and Readable being a superclass should mean any type checks continue to work, this can be implemented as a non-breaking change.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201885737791795/1202445230557606) by [Unito](https://www.unito.io)

### DIFF
--- a/src/httpRequest.ts
+++ b/src/httpRequest.ts
@@ -1,4 +1,4 @@
-import { ReadStream } from "fs";
+import { Readable } from "stream";
 import { request, RequestOptions } from "https";
 import { userAgent } from "./userAgent";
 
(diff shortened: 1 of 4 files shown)


briancbarrow (approved):
Looks good @wdomburg. I'm going to have someone else look at it too before we merge.

briancbarrow (changes requested):
@wdomburg looks like we need to have the test updated too. Will you update the test file `preRecordedTranscription.test.ts` to assert the error message matches `"DG: Mimetype must be provided if the source is a Buffer or a Readable"` That should be on line 16 of that file.

briancbarrow (approved):

PR #60, opened 2022-09-02
added detect language option to preRecordedOptions

briancbarrow:

PR #62, opened 2022-09-02
Adding paragraphs feature to SDK

MichaelJolley:


briancbarrow (approved):
This will affect [this PR](https://github.com/deepgram/deepgram-node-sdk/pull/60) and we might need to change it. I like the refactoring done here though so it probably makes more sense to change PR #60

PR #66, opened 2022-09-19
add detect_topics to pre-recorded options

briancbarrow:


src/types/topic.ts
```diff
@@ -0,0 +1,11 @@
+export type TopicGroup = {
+  topics: Array<Topic>;
```
MichaelJolley, line 2: Need to add property comments.

src/types/topic.ts
```diff
@@ -0,0 +1,11 @@
...
+  text: string;
+  start_word: number;
+  end_word: number;
+};
+
+export type Topic = {
+  topic: string;
```
MichaelJolley, line 9: Need to add property comments.

MichaelJolley (changes requested):
Also need to update the /usage endpoints that allow specifying features

PR #67, opened 2022-09-19
added entity detection

briancbarrow:


src/types/entity.ts
```diff
@@ -0,0 +1,7 @@
+export type Entity = {
+  label: string;
```
MichaelJolley, line 2: Need comments on the properties. I asked Pankaj for these in the .NET ticket.

src/types/prerecordedTranscriptionOptions.ts
```diff
@@ -160,4 +160,9 @@ export type PrerecordedTranscriptionOptions = {
    * diarize, or multichannel to true.
    */
   paragraphs?: boolean;
+
+  /**
+   * Indicates whether Deepgram will Identify and detect entities in the transcript
+   */
```
MichaelJolley, line 166: Is entity detection for pre-recorded only?

MichaelJolley (approved):

PR #79, opened 2022-10-04
updates `keys.test.ts` to make code style consistent across all test files

jgbowser:
Hello, I've chosen to bypass creating an issue for this because I believe it is a pretty trivial test change.

I noticed in `CONTRIBUTING.md` that one of the requirements for contributions to be accepted is that they follow existing styling conventions in the code base. While working through a separate ticket I noticed there was some inconsistency in the styling of callbacks for `describe`, `it`, etc functions. Most test files used arrow functions but some tests in the `keys.test.ts` file used function declarations (some functions in the same file even used arrow functions as well). 

I know this is kind of a nitpick, so feel free to push back if you don't believe this is necessary. but I thought changing everything to be consistent across all test files may help future contributors ensure they are following a consistent code style.

### DIFF
--- a/tests/keys.test.ts
+++ b/tests/keys.test.ts
@@ -19,18 +19,18 @@ describe("Key tests", () => {
   let requestStub: Sinon.SinonStub;
   let keys: Keys;
 
-  beforeEach(function () {
+  beforeEach(() => {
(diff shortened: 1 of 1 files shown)


MichaelJolley (approved):
I agree. Nice cleanup!

PR #112, opened 2023-03-13
added reason to close event

briancbarrow:


src/transcription/liveTranscription.ts
```diff
@@ -36,6 +36,9 @@ export class LiveTranscription extends EventEmitter {
     };
 
     this._socket.onclose = (event: WebSocket.CloseEvent) => {
+      const newTarget: any = event.target;
+      const dgErrorIndex = newTarget["_req"].res.rawHeaders.indexOf("dg-error");
+      event.reason = newTarget["_req"].res.rawHeaders[dgErrorIndex + 1];
```
BrianHillis-DG, line 41: I felt like these had a chance to fail(with the help of TS), so I did some testing. 

I was able to artificially trigger some errors by using an invalid URL. `newTarget["_req"]` is `null` on a 404, resulting in the following: `TypeError: Cannot read property 'res' of null`. Not sure if there are other scenarios in which the `_req` field is `null`, but maybe worth some checks in here to prevent error's in that sphere

Maybe there's a reason TS is angry about the signature `newTarget` as a `WebSocket` :smile:

briancbarrow: Interesting. Maybe I'll set up a conditional check and assign it if it is available?

BrianHillis-DG (approved):
LGTM! 

I would add a comment about the type signature of `const newTarget: any`  to note why it opts for the generic, but totally up to you

PR #127, opened 2023-05-03
added error response

### LINKED ISSUE #117: API errors are not part of the response class declaration

If an API error comes back, we're not typing them in the response.

## Steps to reproduce

Use an API key with no credits to try and transcribe a file. The response class is returned as such

```
PrerecordedTranscriptionResponse {
  err_code: 'ASR_PAYMENT_REQUIRED',
  err_msg: 'Project does not have enough credits for an ASR request and does not have an overage agreement.',
  request_id: 'xxxx'
}
```

But it is defined like this:

```ts
export declare class PrerecordedTranscriptionResponse {
    request_id?: string;
    metadata?: Metadata;
    results?: {
        channels: Array<Channel>;
        utterances?: Array<Utterance>;
    };
    /**
     * Converts the transcription to the WebVTT format
     * @remarks In order to translate the transcription to WebVTT, the utterances
     * feature must be used.
     * @returns A string with the transcription in the WebVTT format
     */
    toWebVTT(): string;
    /**
     * Converts the transcription to the SRT format
     * @remarks In order to translate the transcription to SRT, the utterances
     * feature must be used.
     * @returns A string with the transcription in the SRT format
     */
    toSRT(): string;
}
```

The result is an error when trying to access the error properties of a response.

```ts
    const response = await deepgram.transcription
        .preRecorded(audioSource)
        .catch((err: any) => this.error(err));

    console.log(response.err_msg);
```

`src/commands/transcribe-file/index.ts:98:20 - error TS2339: Property 'err_msg' does not exist on type 'PrerecordedTranscriptionResponse'.`

## Expected behavior

We should reflect all API response properties in our class declarations.

.

## Other information

To work around this locally, you can extend the type like so:

```ts
import { PrerecordedTranscriptionResponse as PrerecordedTranscriptionBase } from "@deepgram/sdk/dist/types";

class PrerecordedTranscriptionResponse extends PrerecordedTranscriptionBase {
  err_code?: string;
  err_msg?: string;
} 

//...

    const response: PrerecordedTranscriptionResponse = await deepgram.transcription
        .preRecorded(audioSource)
        .catch((err: any) => this.error(err));

    console.log(response.err_msg);
```

briancbarrow:
closes #117


lukeocodes (approved):

lukeocodes (changes requested):
This is failing on tests at the moment

lukeocodes (approved):

PR #129, opened 2023-05-15
updated live transcription section in README

briancbarrow:


lukeocodes (approved):

lukeocodes:
This is okay for now because it is removing the browser dependencies of a node-SDK request, but I think a small example of each method in the SDK need to be on our READMEs

PR #131, opened 2023-05-24
Support arbitrary properties in request

lukeocodes:
CFE identified a need to allow certain (will remain nameless) properties to be passed through to the API while avoiding baking in those properties to our type definitions.


bd-g (approved):
Wasn't aware of this Typescript feature - very clean, and meets the requirement we discussed.

briancbarrow (approved):

PR #132, opened 2023-05-24
exposed LiveTranscription for Typescript

briancbarrow:


lukeocodes (changes requested):
makes sense to be able to export this class, but lets export it from `src/transcription/index.ts` rather than a types index.

exporting this will actually help us with something too

briancbarrow:
closing because it is already exported from `"@deepgram/sdk/dist/transcription/liveTranscription";`

PR #136, opened 2023-06-06
Use nullish coalescing

nickottocode:
Updates to the Deepgram constructor to use `??` instead of `||`.  This change makes the SDK compatible with onprem for streaming (batch needs additional updates).  The `apiUrl` could probably be left as `||`, but `requireSSL` needs to be the "nullish coalescing" operator `??` as otherwise if set to "False" for the onprem case the `||` operator will still default to the default option.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -14,6 +14,7 @@ import { _request } from "./httpRequest";
 export class Deepgram {
   private _apiUrl: string;
   private _apiKey: string;
+  private _port: number;
   private _requireSSL: boolean;
 
   keys: Keys;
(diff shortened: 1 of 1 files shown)


lukeocodes (approved):
I have a couple questions for Slack but it LGTM 

for reference and posterity: https://mariusschulz.com/blog/nullish-coalescing-the-operator-in-typescript#providing-fallback-values-with-the-operator

lukeocodes:
@briancbarrow could we tag this as a pre-release before it goes into main and try it on all of the current projects being built?

PR #143, opened 2023-06-16
feat: add support for new beta release of summarisation

briancbarrow:


src/types/prerecordedTranscriptionResponse.ts
```diff
@@ -11,6 +11,9 @@ export class PrerecordedTranscriptionResponse {
     channels: Array<Channel>;
     utterances?: Array<Utterance>;
   };
+  summary?: {
+    short: string;
+  };
```
lukeocodes, line 16: I think this goes inside `results` ? :/

lukeocodes (changes requested):
I think the summary is returned from inside the result? `results.summary.short` is how I had it returning when I used it

lukeocodes (approved):

PR #154, opened 2023-07-18
feat!: Node SDK V3

lukeocodes:


.github/workflows/CI.yml
```diff
@@ -1,30 +1,48 @@
...
+    paths-ignore:
+      - "docs/**"
+      - "**/*.md"
+      - ".prettierrc"
+      - "**/*ignore"
+  push:
+    branches:
+      - main
+      - next
+      - rc
```
SandraRodgers, line 14: Why do we have a 'next' and 'rc' branch?

src/lib/version.ts
```diff
@@ -0,0 +1 @@
+export const version = "0.0.0-automated";
```
SandraRodgers, line 1: How does this version number get changed/automated?

src/packages/PreRecordedClient.ts
```diff
@@ -0,0 +1,174 @@
...
+
+    return body;
+  }
+
+  private async _makeSyncPrerecordedRequest(
+    options: PrerecordedOptions | undefined,
+    endpoint: string,
+    body: string | Buffer | Readable
+  ): Promise<DeepgramResponse<SyncPrerecordedResponse>> {
+    if (options !== undefined && "callback" in options) {
```
SandraRodgers, line 135: It's possible they would add a callback to the request like this:

```
deepgram.listen.prerecorded.transcribeUrl(audio_url, options, callback)
```

This wouldn't cause the error to throw because it would assume the callback url is the endpoint, right? So it would show them a 400 error instead of this one.

lukeocodes: We're going to accept an object as the callback URL

src/lib/types/CreateProjectKeyOptions.ts
```diff
@@ -0,0 +1,16 @@
+export type CreateProjectKeyOptions = ExpirationOptions | TtlOptions;
+
+interface ExpirationOptions extends CommonOptions {
+  expiration_date: string;
+}
+
+export interface TtlOptions extends CommonOptions {
+  time_to_live_in_seconds: string;
```
SandraRodgers, line 8: I think the API expects an integer for this, not a string

.github/workflows/CI.yml
lukeocodes: I am considering introducing them as our formal release pipeline. Essentially you create a `next` branch from an `rc`, and an `rc` on a tag which you believe is ready to be released but needs to be tested in the wild. They imply levels of stability.

src/packages/PreRecordedClient.ts
```diff
@@ -0,0 +1,188 @@
...
+        body = source.buffer;
+      } else if (isReadStreamSource(source)) {
+        body = source.stream;
+      } else {
+        throw new DeepgramError("Unknown transcription source type");
+      }
+
+      if (options !== undefined && "callback" in options) {
+        throw new DeepgramError(
+          "Callback cannot be provided as an option to a synchronous transcription. Use `asyncPrerecordedUrl` or `asyncPrerecordedFile` instead."
```
SandraRodgers, line 84: Need to update comment now that aren't using name 'asyncPrerecordedUrl'

PR #172, opened 2023-10-12
feat: remove duplicate example apps

lukeocodes:
## Proposed changes

Removes the duplicate examples left over from the V2 node SDK. Both live and prerecorded examples now use V3 JS SDK


jpvajda (approved):
LGTM

PR #177, opened 2023-10-17
feat: remove mimetype requirement from batch requests (rfc #4)

lukeocodes:


dvonthenen (commented):
Taking a look at this and got a question on the change. Is the mimetype optional since it's being removed? Or is the mimetype being inferred by the file type in batch operations?

dvonthenen (approved):

PR #180, opened 2023-10-19
feat: making schema's possible from request types

lukeocodes:


dvonthenen (approved):
The "Options" name is very Go-esque. Cross of styles from the original author?

lukeocodes:
> The "Options" name is very Go-esque. Cross of styles from the original author?

Kinda, I think a lot of it was left over from when I based it on the supabase-js package.

PR #183, opened 2023-10-24
fix(types): nested arrays in LiveTranscriptionEvent

ftr-lwilson:
The `LiveTranscriptionEvent` type specifies that the `alternatives` and `words` nested fields are tuples of length 1, when they should be variable length arrays.

image


lukeocodes:
Oh, this is a great catch! I'll have a scan through today, there could be more instances of this

lukeocodes (approved):
lgtm. merging to a branch so i can catch any other examples

PR #245, opened 2024-02-02
feat: add fetch override

### LINKED ISSUE #243: Increase in `UND_ERR_HEADERS_TIMEOUT` errors
In periods of high load, response times fluctuate. If they go over 5' mark, the fetch library implemented fails, because it cannot stay alive for longer than 5'. Those timeouts are also not configurable.

This is a [known limitation](https://github.com/nodejs/node/issues/46375) of the fetch library implemented.

## Steps to reproduce
It's ad-hoc and depends on response rates & usage of Deepgram's servers

## Expected behavior
That we don't see timeout issues and can configure timeout values if needed.

- **Operating System/Version:** Mac
- **Language:** [TypeScript]
- **Browser:** Chrome

This is from calls made server side

## Other information
Screenshot 2024-01-30 at 9 04 32 AM

BartoszJarocki:
This PR renames `fetch` to `fetchOptions` and adds an additional option to provide own fetch method. 

fixes #243

### DIFF
--- a/src/lib/constants.ts
+++ b/src/lib/constants.ts
@@ -1,4 +1,5 @@
 import { isBrowser } from "./helpers";
+import { DeepgramClientOptions } from "./types/DeepgramClientOptions";
 import { FetchOptions } from "./types/Fetch";
 import { version } from "./version";
 
@@ -18,7 +19,7 @@ export const DEFAULT_FETCH_OPTIONS: FetchOptions = {
   headers: DEFAULT_HEADERS,
 };
 
-export const DEFAULT_OPTIONS = {
+export const DEFAULT_OPTIONS: DeepgramClientOptions = {
   global: DEFAULT_GLOBAL_OPTIONS,
-  fetch: DEFAULT_FETCH_OPTIONS,
+  fetchOptions: DEFAULT_FETCH_OPTIONS,
 };
--- a/src/lib/constants.ts
+++ b/src/lib/constants.ts
@@ -21,5 +21,5 @@ export const DEFAULT_FETCH_OPTIONS: FetchOptions = {
 
 export const DEFAULT_OPTIONS: DeepgramClientOptions = {
   global: DEFAULT_GLOBAL_OPTIONS,
-  fetchOptions: DEFAULT_FETCH_OPTIONS,
+  fetch: DEFAULT_FETCH_OPTIONS,
 };
--- a/src/lib/fetch.ts
+++ b/src/lib/fetch.ts
@@ -2,9 +2,11 @@ import crossFetch from "cross-fetch";
 import { resolveHeadersConstructor } from "./helpers";
 import type { Fetch } from "./types/Fetch";
 
-export const resolveFetch = (): Fetch => {
+export const resolveFetch = (customFetch?: Fetch): Fetch => {
   let _fetch: Fetch;
-  if (typeof fetch === "undefined") {
+  if (customFetch) {
+    _fetch = customFetch;
+  } else if (typeof fetch === "undefined") {
     _fetch = crossFetch as unknown as Fetch;
   } else {
     _fetch = fetch;
@@ -12,8 +14,8 @@ export const resolveFetch = (): Fetch => {
   return (...args) => _fetch(...args);
 };
 
-export const fetchWithAuth = (apiKey: string): Fetch => {
-  const fetch = resolveFetch();
+export const fetchWithAuth = (apiKey: string, customFetch?: Fetch): Fetch => {
+  const fetch = resolveFetch(customFetch);
   const HeadersConstructor = resolveHeadersConstructor();
 
   return async (input, init) => {
--- a/src/lib/types/DeepgramClientOptions.ts
+++ b/src/lib/types/DeepgramClientOptions.ts
@@ -13,11 +13,8 @@ export interface DeepgramClientOptions {
     url?: string;
   };
   fetchOptions?: FetchOptions;
+  fetch?: Fetch;
   restProxy?: {
     url: null | string;
   };
-}
-
-export interface DeepgramClientOptionsWithFetchOverride extends DeepgramClientOptions {
-  fetch: Fetch;
 }
--- a/src/lib/types/DeepgramClientOptions.ts
+++ b/src/lib/types/DeepgramClientOptions.ts
@@ -1,4 +1,4 @@
-import { FetchOptions } from "./Fetch";
+import { Fetch, FetchOptions } from "./Fetch";
 
 export interface DeepgramClientOptions {
   global?: {
@@ -12,8 +12,12 @@ export interface DeepgramClientOptions {
      */
     url?: string;
   };
-  fetch?: FetchOptions;
+  fetchOptions?: FetchOptions;
   restProxy?: {
     url: null | string;
   };
+}
+
+export interface DeepgramClientOptionsWithFetchOverride extends DeepgramClientOptions {
+  fetch: Fetch;
 }
--- a/src/lib/types/DeepgramClientOptions.ts
+++ b/src/lib/types/DeepgramClientOptions.ts
@@ -12,8 +12,8 @@ export interface DeepgramClientOptions {
      */
     url?: string;
   };
-  fetchOptions?: FetchOptions;
-  fetch?: Fetch;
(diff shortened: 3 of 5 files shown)


lukeocodes (changes requested):
Thanks, I really appreciate the effort here.

The only issue with this is that it is a breaking change and we're not ready to add another major version to our release channel. I would encourage you to add a new property (not breaking) like `fetchClient`. 

I would then replace `fetchWithAuth` with a function that switched Fetch if it was supplied, and then called fetchWithAuth to return a preconfigured fetch instance using the right class and the auth headers. 

Superbase has a good example, and they use cross-fetch too https://github.com/supabase/supabase-js/blob/master/src/lib/fetch.ts#L6

Also any tests you're able to provide would be very helpful

dvonthenen (changes requested):
Clicked approve on the wrong PR. Should have been this one:
https://github.com/deepgram/deepgram-js-sdk/pull/246#pullrequestreview-1863019864

BartoszJarocki:
@lukeocodes understood, my thinking was that it'll be a bit confusing to have `fetch` as fetch options and something different to actually override fetch. now, I changed the implementation to be similar to supabase example, thank you!  would you mind taking a look again? 

regarding tests, I personally am not a fan of testing protected methods so the only option I see is to add tests like that to each client type

```
  it("should use custom fetch when provided", async () => {
    const fetch = async () => {
      return new Response(JSON.stringify({ customFetch: true }));
    };
 
    const client = createClient(faker.string.alphanumeric(40), {
      global: { url: "https://api.mock.deepgram.com" },
      customFetch: fetch,
    });
 
    const { result, error } = await client.manage.getProjectBalances(faker.string.uuid());
 
    assert.isNull(error);
    assert.isNotNull(result);
    assert.containsAllDeepKeys(result, ["customFetch"]);
  });
```

wdyt?

lukeocodes:
Looks good to me! I'm going to personally test this on browsers, edge functions and Node environments to ensure we're not introducing a regression. I'll try and get this reviewed by the end of the week. Really appreciate you coming back with fixes for the PR

pkpio:
Any update on this? Would be great to also see an example somewhere of a custom fetch impl to use - trying to workaround the 5 mins timeout errors of the default nodejs fetch

BartoszJarocki:
@pkpio I added some examples here https://github.com/BartoszJarocki/deepgram-js-sdk-custom-fetch/blob/main/index.ts

myselfhimself:
> @lukeocodes understood, my thinking was that it'll be a bit confusing to have `fetch` as fetch options and something different to actually override fetch. now, I changed the implementation to be similar to supabase example, thank you! would you mind taking a look again?
> 
> regarding tests, I personally am not a fan of testing protected methods so the only option I see is to add tests like that to each client type
> 
> ```
>   it("should use custom fetch when provided", async () => {
>     const fetch = async () => {
>       return new Response(JSON.stringify({ customFetch: true }));
>     };
>  
>     const client = createClient(faker.string.alphanumeric(40), {
>       global: { url: "https://api.mock.deepgram.com" },
>       customFetch: fetch,
>     });
>  
>     const { result, error } = await client.manage.getProjectBalances(faker.string.uuid());
>  
>     assert.isNull(error);
>     assert.isNotNull(result);
>     assert.containsAllDeepKeys(result, ["customFetch"]);
>   });
> ```
> 
> wdyt?

Your test looks good, it could be added to https://github.com/deepgram/deepgram-js-sdk/blob/main/test/client.test.ts if I am not mistaken..

myselfhimself:
Hello here is some example working code using an AbortController `.signal`-enabled custom `fetch` with @BartoszJarocki 's pull request:

```
import { DeepgramError, createClient } from "@deepgram/sdk";
import { SettingsManagerCookies as SettingsManager } from "./settingsManager";

export class DeepgramClientCancellationError extends DeepgramError {
    originalError: unknown;

    constructor(message: string, originalError: unknown) {
        super(message);
        this.name = "DeepgramClientCancellationError";
        this.originalError = originalError;
    }
}

export const createDeepgramClient = (apiKey?: string, signal?: AbortSignal) => {
    async function cancellableFetch(
        input: RequestInfo | URL, init?: RequestInit
    ): Promise<Response> {
        // Use cross-fetch with the AbortController signal
        try {
            const response = await fetch(input, { ...init, signal });
            console.log("got response in custom Fetch, with signal" + JSON.stringify(signal));
            return response;
        } catch (error) {
            if ((error as DOMException).name === "AbortError") {
                // Handle fetch abort due to timeout
                throw new DeepgramClientCancellationError("Deepgram query aborted by user", error);
            } else {
                // Re-throw other errors
                throw error;
            }
        }
    }



    var proxyUrl = "http://127.0.0.1:8889";

    console.log("creating client with Deepgram experimental customFetch parameter");
    return createClient("proxy", {
        restProxy: { url: proxyUrl },
        fetch: { headers: { forceApiKey: apiKey ?? (SettingsManager.getApiKey()) } },
        customFetch: cancellableFetch
    });
}

// Example usage:
var controller = new AbortController();
var client = createDeepgramClient(undefined /* using my settings manager */, controller.signal);
client.listen.prerecorded.transcribeFile(binFile, options).then((result) => {}).catch((err) => { if(err.name == "AbortController") { console.log("was cancelled"); })
// Some cancelling somewhere, on some click() or interrupt.
controller.abort();
```

To use the custom fetch pull request, I used the following `package.json` line: `    "@deepgram/sdk": "github:deepgram/deepgram-js-sdk#pull/245", ` and I had to `cd` into `node_modules/deepgram/sdk` and run `npm i` followed by `npm run build`.
  
The forceApikey mechanism allows a deepgram client to tell a deepgram proxy to use a different API key, it is described in the [deepgram http proxy project.](https://github.com/deepgram-devs/deepgram-node-proxy/issues/2), though no deepgram proxy implementation has been published yet IMHO.

myselfhimself:
I am about to release a product with this fork...
Here are some information for people wanting to use the fork  with package.json in a CI/CD pipeline:
package.json:
```
...
"scripts": {
"prebuild": "ls node_modules/@deepgram/sdk | grep -q tsconfig.json && npm i --prefix node_modules/@deepgram/sdk && npm run build --prefix node_modules/@deepgram/sdk
....}
...
"dependencies": {
...
    "@deepgram/sdk": "https://github.com/BartoszJarocki/deepgram-js-sdk/tarball/feat/add-fetch-override",
...
}
...
```

lukeocodes:
@BartoszJarocki thanks for your patience on this. If you could please add the test just as you had it, that will be great.

FYI, I have modified the namespace of the custom fetch to `_experimentalCustomFetch`. I am keen not to take up any more namespaces before I figure out how I want to allow settings to differ between products on the same instance. We have some thinking to do (as we've added more products to the SDK).

Once we have the test, this is good to go. You can add it to `client.test.ts`. Thanks again for your patience and help.

bjarockifilmio:
@lukeocodes I'm glad it's good to go! I just added a test [84a4e27](https://github.com/deepgram/deepgram-js-sdk/pull/245/commits/84a4e2725dc68bca353a165cd58c4c69678e1cbe).

BartoszJarocki:
@lukeocodes I'm glad it's good to go! I just added a test [84a4e27](https://github.com/deepgram/deepgram-js-sdk/pull/245/commits/84a4e2725dc68bca353a165cd58c4c69678e1cbe).

lukeocodes:
Sweet. I'll just double check some stuff, might cut a release today all
being well and having a colleague available to check a box 😇

Thanks again for your help and patience! I appreciate it's been sitting
here for a month or two, while we we cleared our plates!

BartoszJarocki:
awesome, sounds good!

myselfhimself:
Now using the 3.3.0 version, with a unit test to ensure the experimental custom fetch parameter keeps working before releasing myself. Thank you @BartoszJarocki @lukeocodes !!

lukeocodes:
Thanks again!

I have now (now that I have some time) been looking at a long term solution for transmitter overrides (websocket and fetch), custom headers, namespaced config per product (transcription and speech intelligence can operate differently), etc, etc. 

I have a good solution that is backwards compatible, and will maintain compatibility with the experimental property we included here.

Mind if I include you in the review of it when it is ready?

myselfhimself:
Good!

PR #256, opened 2024-03-16
fix: add missing paragraphs field to Result in SyncPrerecordedResponse

dbolotin:


lukeocodes:
What is this change for?

lukeocodes:
Paragraphs are not returned as part of the `result` object, they belong in the alternatives children.

PR #264, opened 2024-04-16
Handle is_final and endpointing together with utterance end

DamienDeepgram:
## Proposed changes
This is an improved version to make it easier to understand the docs [here](https://developers.deepgram.com/docs/understand-endpointing-interim-results)

Users tend to not understand how to handle these things and our samples do not include how it works

## Types of changes

What types of changes does your code introduce to the community Python SDK?
_Put an `x` in the boxes that apply_

- [x] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update or tests (if none of the other choices apply)

### DIFF
--- a/examples/node-live/index.js
+++ b/examples/node-live/index.js
@@ -1,8 +1,4 @@
-const {
-  createClient,
-  LiveTranscriptionEvents,
-  LiveTranscriptionEvent,
-} = require("../../dist/main/index");
+const { createClient, LiveTranscriptionEvents } = require("../../dist/main/index");
 const fetch = require("cross-fetch");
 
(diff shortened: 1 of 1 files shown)


lukeocodes:
hey bud, what's this for? might be worth making this a separate example file. don't be afraid to make as many examples as you want, just name them appropriately

DamienDeepgram:
This is to properly handle the websocket messages for endpointing, utterance end and interim results

lukeocodes (approved):

PR #269, opened 2024-04-25
fix: update user agent to include JS or node

lukeocodes:
adding sdk to user agent


dvonthenen (approved):

jpvajda (approved):
Thank you for making this change.

PR #270, opened 2024-04-25
feat: reconnect socket automatically on close event

lukeocodes:


src/packages/LiveClient.ts
```diff
@@ -16,33 +16,62 @@ import type {
...
+  }
 
-    this._socket = new w3cwebsocket(url.toString(), ["token", this.key]);
+  protected _attemptReconnection(): void {
+    if (this._reconnections > WS_RECONNECTION_RETRY_LIMIT) {
+      throw new DeepgramError("Reconnection retry limit exceeded");
+    }
+
+    this._reconnections += 1;
+    this._unregisterEvents();
```
dvonthenen, line 52: Would the user be interested in getting the `onopen` event to know that the WS successfully reconnected?

test/live.test.ts
```diff
@@ -7,9 +7,9 @@ import { LiveConnectionState, LiveTranscriptionEvents } from "../src/lib/enums";
 describe("connecting to our transcription websocket", () => {
   let deepgram: DeepgramClient;
 
-  beforeEach(() => {
+  before(() => {
     deepgram = createClient(faker.string.alphanumeric(40), {
-      global: { url: "https://api.mock.deepgram.com" },
+      global: { url: "https://api.mock.deepgram" },
```
dvonthenen, line 12: Not sure if the removal of `.com` is intentional.

dvonthenen (commented):
Just checking to see if these are desired or not.

test/live.test.ts
lukeocodes: It is intentional! But the commit of it wasn't lol. thanks!

lukeocodes: fixed

src/packages/LiveClient.ts
lukeocodes: We reregister the events after. These internal events emit an external one, which won't need to be reregistered. So they're likely to get second onopen event... which may be troublesome for some actually

lukeocodes:
I need to get the key factory in and documented before this, and I may change this to be a method users can call rather than automatically reconnect

PR #273, opened 2024-04-29
feat: namespace configs for transmitters and options

lukeocodes:
Full compatible with existing config (through a legacy config transformer)

The client options will look like this for custom headers for Fetch and WS Node clients;

```js
{
  global: {
    fetch: { options: { headers: { "customheader": "value" } } },
    websocket: { options: { _nodeOnlyHeaders: { "customheader": "value" } } }
  }
  // ... other namespaces
} 
```

Namespace configurations are coming, too.

You'll now be able to provide config to specific namespaces in the SDK.

```js
{
  global: {
    fetch: { options: { headers: { "customheader": "value" } } },
  },
  listen: {
    fetch: { options: { 
      url: "http://localhost:3000"
    } },
  }
  // ... other namespaces
} 
```

Finally, for the holy grail of overriding a client, which will also be available in namespace config, and will have access to the namespaceOptions scope.

```js
{
  listen: {
    client: async (options) => {} // their client here
  }
} 
```


dvonthenen (approved):

dennisofficial:
Oooooo. Elaborate the namespace? You mean we can abstract the client classes? Like 'PrerecordedClient' class?

lukeocodes:
You'll be able to configure listen separately of read, speak etc from the
client options. Examples in the PR. using the classes directly is possible
but risky, as it's not a public interface it can change without a major
version increase

dennisofficial:
Does this allow us to set a fetch timeout or abort signal? That would be so helpful ❤️

dennisofficial:
The only reason I would override the class is to include an abort signal code so I can attach the signal property to the fetch request.

PR #275, opened 2024-04-30
fix: process is undefined error in browsers

### LINKED ISSUE #274: v3.3.0 Error in the browser: `Uncaught ReferenceError: process is not defined`
Installing v3.3.0 results in the error `Uncaught ReferenceError: process is not defined`

## What's happening that seems wrong?
It seems that the dist bundle is not correct. dist bundles should have references to `process` object

## Steps to reproduce
1. Install deepgram-sdk and use it in a browser application
2. Run the application
3. Visit the webpage

## Expected behavior
1. The application should simply run
2. And you can use the deepgram client.

- **Operating System/Version:** MacOS
- **Language:** TypeScript
- **Browser:** Chrome

## Other information
Reverting to v3.2.0 resolves the issue.

### Anything else we should know? (e.g. detailed explanation, stack-traces, related issues, suggestions how to fix, links for us to have context, eg. stack overflow, codepen, etc)
1. deepgram v3.2 Example: https://stackblitz.com/edit/vitejs-vite-fhctlf?file=package.json (_This works fine_)
2. deepgram v3.3 Example: https://stackblitz.com/edit/vitejs-vite-zg6vpu?file=package.json (_This fails_)

image

lukeocodes:
fixes #274 

writing the node version at build to the browser scope, as we don't report on it anyway. nice quick fix.

### DIFF
--- a/examples/browser-prerecorded/index.html
+++ b/examples/browser-prerecorded/index.html
@@ -1,0 +1,31 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <script src="../../dist/umd/deepgram.js"></script>
+    <script>
+      const { createClient } = deepgram;
+      const _deepgram = createClient("deepgram-api-key", {
+        global: {
+          fetch: {
+            options: {
+              url: "https://api.mock.deepgram.com",
+            },
+          },
+        },
+      });
+
+      console.log("Deepgram Instance: ", _deepgram);
+
+      (async () => {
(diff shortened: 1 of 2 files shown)


SandraRodgers (approved):

alabhyajindal:
Hi! I'm facing the same issue right now. I have a SvelteKit app. When I import the SDK:

```ts
<script lang="ts">
	import { createClient } from '@deepgram/sdk';
</script>
```

I get the following error:
image

PR #276, opened 2024-05-02
chore: simplify regex

msopacua:
- We already do toLowerCase(), so don't need `i` flag.
- We're not using the http bit, so no need to capture it.
- If there's an s, append it, if not don't. This is the same as just replacing http with ws:
  - [http]:// -> [ws]://
  - [http]s:// -> [ws]s://
  - [http]x:// -> [ws]x:// -> error. Is ok.
  - [http]sx:// -> [ws]sx:// -> error in both. Is ok.
- We don't want to replace protocols that do not start with http, such as git+https, so we can anchor at string start and leave out the `g` flag.

Much simpler -:)

### DIFF
--- a/src/packages/LiveClient.ts
+++ b/src/packages/LiveClient.ts
@@ -27,7 +27,7 @@ export class LiveClient extends AbstractWsClient {
     super(key, options);
 
     const url = new URL(endpoint, this.baseUrl);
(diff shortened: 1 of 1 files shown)


lukeocodes:
Oh it's possible this is different as of #273, I can't remember if I replaced this code or moved it

lukeocodes (approved):
approved - but will wait for 3.4

PR #279, opened 2024-05-03
process is undefined error vite

rens-public:
This pull request addresses an issue with the Deepgram SDK when used in Vite environments.

Currently, the SDK relies on the process.versions.node property to determine the Node.js version. However, in Vite environments, the process object is not available, leading to errors.

The proposed fix is a check to verify the existence of the process object before accessing process.versions.node. If process is undefined, the NODE_VERSION constant defaults to "unknown".

This change ensures compatibility with Vite and other environments where the process object is not accessible.

### DIFF
--- a/src/lib/constants.ts
+++ b/src/lib/constants.ts
@@ -3,7 +3,10 @@ import { DeepgramClientOptions } from "./types/DeepgramClientOptions";
 import { FetchOptions } from "./types/Fetch";
 import { version } from "./version";
 
-export const NODE_VERSION = process.versions.node;
+export const NODE_VERSION =
+  typeof process !== "undefined" && process.versions && process.versions.node
+    ? process.versions.node
+    : "unknown";
 
 export const DEFAULT_HEADERS = {
   "Content-Type": `application/json`,
--- a/src/lib/helpers.ts
+++ b/src/lib/helpers.ts
@@ -82,6 +82,4 @@ const isReadStreamSource = (providedSource: PrerecordedSource): providedSource i
   return false;
 };
 
-export class CallbackUrl extends URL {
-  private callbackUrl = true;
-}
+export class CallbackUrl extends URL {}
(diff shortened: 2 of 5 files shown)


lukeocodes:
I think it would be better to use `vite-plugin-env-compatible` rather than deliberately break something. Do you want to make a new PR? Or reset this back a few commits and redo it?

Something like this?

```
  plugins: [
    envCompatible({
      process: true, // This enables the `process` global
    }),
  ],
  define: {
    'process.versions.node': JSON.stringify(process.versions.node),
  }
```

src/lib/constants.ts
```diff
@@ -3,7 +3,10 @@ import { DeepgramClientOptions } from "./types/DeepgramClientOptions";
 import { FetchOptions } from "./types/Fetch";
 import { version } from "./version";
 
-export const NODE_VERSION = process.versions.node;
+export const NODE_VERSION =
```
lukeocodes, line 6: This will potentially affect how we trace SDK version usage on our API. You should look into the vite process plugin to enable process.versions.node instead.

src/lib/helpers.ts
```diff
@@ -82,6 +82,4 @@ const isReadStreamSource = (providedSource: PrerecordedSource): providedSource i
   return false;
 };
 
-export class CallbackUrl extends URL {
-  private callbackUrl = true;
-}
+export class CallbackUrl extends URL {}
```
lukeocodes, line 85: Not sure why you have done this.

src/packages/SpeakClient.ts
```diff
@@ -1,5 +1,5 @@
 import { AbstractRestfulClient } from "./AbstractRestfulClient";
-import { DeepgramError, DeepgramUnknownError, isDeepgramError } from "../lib/errors";
+import { DeepgramError, DeepgramUnknownError } from "../lib/errors";
```
lukeocodes, line 2: A little advice would be to keep your PRs atomic, and change only things relevant to what you want changed.

lukeocodes (changes requested):

src/lib/constants.ts
noxan: @lukeocodes What's the point of tracing a nodejs version from an environment which does not even have nodejs (only during the build step)? It would make a lot more sense to look at the browser version and environment than nodejs.

lukeocodes: you're assuming that this only runs in the browser

PR #292, opened 2024-05-20
fix: unconditional process reference

### LINKED ISSUE #290: TypeError: Cannot read property 'node' of undefined
After upgrading from 3.2.0 to 3.3.2 I'm running into the error `TypeError: Cannot read property 'node' of undefined`. I'm experiencing this in a non-node backend env and inside of a react native env.

msopacua:
The `process` global does not exist in the browser and therefore throws an error if one tries to create a client in the browser.
Fix it in a way that's compatible with any (future) engine that doesn't have a process global.

closes #290


lukeocodes:
Thank you for your contribution!

lukeocodes (approved):

PR #299, opened 2024-05-31
fix: Resolve missing process variable in browser environment

### LINKED ISSUE #297: process is undefined on v3.3.1
There are occations where process is still undefined when using a non-webpack app. we need a solution that doesn't rely on a webpack plugin

noxan:
Follow up to close #297 

- check if `process` is defined, else fall back to `Unknown`
- prevent the unlikely case of `process.versions` being defined without a `node` key, fall back to `Unknown`


noxan:
@lukeocodes added a fix with this pull request

lukeocodes (approved):

lukeocodes:
nice, this is a neater solution than a vite config

PR #306, opened 2024-07-03
feat: add TTS Live Client

SandraRodgers:
Introduce live speak functionality to existing `speak` client. Maintain existing interface for batch.

## Summary by CodeRabbit

- **New Features**
	- Added detailed documentation for using the Deepgram Text-to-Speech API, including REST and WebSocket methods.
	- Introduced live text-to-speech functionality, allowing real-time audio synthesis.
	- Added new classes for managing text-to-speech requests and live connections, enhancing user interaction.

- **Bug Fixes**
	- Simplified export structure for the SpeakRestClient class.

- **Documentation**
	- Updated README with practical examples for both synchronous and asynchronous TTS implementations.
	- Expanded documentation with event handling details for live text-to-speech synthesis.

### DIFF
--- a/README.md
+++ b/README.md
@@ -391,7 +391,21 @@ const { result } = await deepgram.speak.request({ text }, { model: "aura-asteria
 ## Websocket
 
 ```js
-const connection = deepgram.speak.live({ model: "aura-asteria-en" });
+const dgConnection = deepgram.speak.live({ model: "aura-asteria-en" });
+
+dgConnection.on(LiveTTSEvents.Open, () => {
+  console.log("Connection opened");
+
+  // Send text data for TTS synthesis
+  dgConnection.sendText(text);
+
+  // Send Flush message to the server after sending the text
+  dgConnection.flush();
+
+  dgConnection.on(LiveTTSEvents.Close, () => {
+    console.log("Connection closed");
+  });
+});
 ```
 
 [See our API reference for more info](https://developers.deepgram.com/reference/text-to-speech-api).
--- a/README.md
+++ b/README.md
@@ -382,8 +382,16 @@ const vttOutput = webvtt(result);
 
 # Text to Speech
 
+## Rest
+
 ```js
 const { result } = await deepgram.speak.request({ text }, { model: "aura-asteria-en" });
+```
+
+## Websocket
+
+```js
+const connection = deepgram.speak.live({ model: "aura-asteria-en" });
 ```
 
 [See our API reference for more info](https://developers.deepgram.com/reference/text-to-speech-api).
--- a/examples/node-speak-live/index.js
+++ b/examples/node-speak-live/index.js
@@ -6,41 +6,41 @@ const live = async () => {
 
   const deepgram = createClient(process.env.DEEPGRAM_API_KEY);
 
-  const connection = deepgram.speak.live({ model: "aura-asteria-en" });
+  const dgConnection = deepgram.speak.live({ model: "aura-asteria-en" });
 
   let audioBuffer = Buffer.alloc(0);
 
-  connection.on(LiveTTSEvents.Open, () => {
+  dgConnection.on(LiveTTSEvents.Open, () => {
     console.log("Connection opened");
 
     // Send text data for TTS synthesis
-    connection.sendText(text);
+    dgConnection.sendText(text);
 
     // Send Flush message to the server after sending the text
-    connection.flush();
+    dgConnection.flush();
 
-    connection.on(LiveTTSEvents.Close, () => {
+    dgConnection.on(LiveTTSEvents.Close, () => {
       console.log("Connection closed");
     });
 
-    connection.on(LiveTTSEvents.Metadata, (data) => {
+    dgConnection.on(LiveTTSEvents.Metadata, (data) => {
       console.dir(data, { depth: null });
     });
 
-    connection.on(LiveTTSEvents.Audio, (data) => {
+    dgConnection.on(LiveTTSEvents.Audio, (data) => {
       console.log("Deepgram audio data received");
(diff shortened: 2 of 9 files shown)


src/packages/AbstractLiveClient.ts
```diff
@@ -258,6 +259,48 @@ export abstract class AbstractLiveClient extends AbstractClient {
...
+    // To be implemented by subclasses
+  }
+
+  /**
+   * Handles binary messages received from the WebSocket connection.
+   * @param data - The binary data.
+   */
+  protected handleBinaryMessage(data: ArrayBuffer): void {
+    // To be implemented by subclasses
+  }
```
lukeocodes, line 303: this appears to be incomplete? what are subclasses?

lukeocodes (changes requested):

examples/node-speak-live/index.js
```diff
@@ -0,0 +1,44 @@
+const { createClient, LiveTTSEvents } = require("../../dist/main/index");
+const fetch = require("cross-fetch");
+
+const live = async () => {
+  const text = "Hello, how can I help you today?";
+
+  const deepgram = createClient(process.env.DEEPGRAM_API_KEY, {
+    global: { fetch: { options: { url: "https://api.beta.deepgram.com" } } },
```
naomi-lgbt, line 8:
```suggestion
    global: { websocket: { options: { url: "wss://api.beta.deepgram.com" } } },
```

It seems to look for a websocket override, not a fetch override.

examples/node-speak-live/index.js
```diff
@@ -0,0 +1,44 @@
...
+  });
+
+  const connection = deepgram.speak.live({ text }, { model: "aura-asteria-en" });
```
naomi-lgbt, line 11:
```suggestion
  const connection = deepgram.speak.live({ model: "aura-asteria-en" });
```

The function definition for this expects `SpeakSchema, endpoint` - passing the `model` object as the second argument broke the function.

src/packages/AbstractLiveClient.ts
```diff
@@ -2,6 +2,7 @@ import { AbstractClient, noop } from "./AbstractClient";
 import { CONNECTION_STATE, SOCKET_STATES } from "../lib/constants";
 import type { DeepgramClientOptions, LiveSchema } from "../lib/types";
```
lukeocodes, line 3: You shouldn't need to make any changes to the Abstract client for this work. These edits should instead live inside `SpeakLiveClient.ts`

src/packages/index.ts
```diff
@@ -7,4 +7,4 @@ export * from "./ListenRestClient";
 export * from "./ManageRestClient";
 export * from "./ReadRestClient";
 export * from "./SelfHostedRestClient";
-export * from "./SpeakRestClient";
```
lukeocodes, line 10: this file is called a barrel file. it should export all clients that exist in the directory. please export `SpeakRestClient`, `SpeakLiveClient`, and `SpeakClient`

src/packages/SpeakClient.ts
```diff
@@ -0,0 +1,32 @@
...
+ * 2. `live(ttsOptions: SpeakSchema = {}, endpoint = ":version/speak")`: Returns a `SpeakLiveClient` instance for interacting with the live speak API, with the provided TTS options and endpoint.
+ */
+export class SpeakClient extends AbstractClient {
+  public namespace: string = "speak";
+
+  /**
+   * Returns a `SpeakRestClient` instance for interacting with the rest speak API.
+   */
+  get request() {
+    return new SpeakRestClient(this.options);
```
lukeocodes, line 20: This is good, I assume to provide backwards compatibility with the dot notation we set up for TTS REST

lukeocodes: however I think for this one, there is a slight gotcha with how we initially set it up, and how to maintain the compatibility of it.

If you make `get request(){}` look like this...

```
  get request(
    source: TextSource,
    options?: SpeakSchema,
    endpoint = ":version/speak"
  ) {
    const client = new SpeakRestClient(this.options);
 
    return client.request(source, options, endpoint);
  }
```

it _should_ maintain the interface that exists now to do `deepgram.speak.request(...etc)`, despite adding a class between so we can also do `deepgram.speak.live(...etc)`

lukeocodes (changes requested):

src/packages/index.ts
lukeocodes: For this, I use a tool called auto-barrel which allows you to right-click on a directory and click on "update barrel file", which does it all for you.

README.md
```diff
@@ -382,10 +382,18 @@ const vttOutput = webvtt(result);
 
 # Text to Speech
 
+## Rest
+
 ```js
 const { result } = await deepgram.speak.request({ text }, { model: "aura-asteria-en" });
 ```
 
+## Websocket
```
lukeocodes, line 391: Can you check out the other websocket example. It's a little more comprehensive than just the connection.

lukeocodes (approved):
Do we want to release this on alpha/beta or straight to main while the API is still in development?

jpvajda:
@lukeocodes  Things we need to probably change in this PR:

* Rename of reset to clear
* Clear message now has a response type.
* We want to remove the container field 

cc @dvonthenen

dvonthenen:
> @lukeocodes Things we need to change in this PR:
> 
> * Rename of reset to clear
> * Clear message now has a response type.
> * We want to remove the container field
> 
> cc @dvonthenen

Yep, confirmed those should be the only changes between now and towards the end of July. 👍

jpvajda (approved):
LGMT!

PR #313, opened 2024-07-09
fix: missing type for detect_language which supports multiple strings OR a single bool

### LINKED ISSUE #295: Support restricting detectable languages
## Proposed changes

> Provide a detailed description of the change or addition you are proposing

The Deepgram API provides the ability to restrict language detection to a subset of all possible languages ([doc link](https://developers.deepgram.com/docs/language-detection#restricting-the-detectable-languages)). Currently, `detect_language` is [only a boolean](https://github.com/deepgram/deepgram-js-sdk/blob/main/src/lib/types/TranscriptionSchema.ts#L154) in this SDK.

The request format is to allow for passing multiple instances of this parameter with string language codes, e.g. `detect_language=en&detect_language=es` (to detect either English or Spanish).

## Context

> Why is this change important to you? How would you use it? How can it benefit other users?

This enables Deepgram users to restrict their languages for language detection, boosting accuracy and improving transcripts.

## Possible Implementation

> Not obligatory, but suggest an idea for implementing addition or change

## Other information

> Anything else we should know? (e.g. detailed explanation, related issues, links for us to have context, eg. stack overflow, codepen, etc)

Deepgrammer opening this issue on behalf of a direct customer request.

lukeocodes:
fixes #295 

adds missing type definition for restricting language detection to an array of language codes

### DIFF
--- a/src/lib/types/TranscriptionSchema.ts
+++ b/src/lib/types/TranscriptionSchema.ts
@@ -149,7 +149,7 @@ interface PrerecordedSchema extends TranscriptionSchema {
   /**
    * @see https://developers.deepgram.com/docs/language-detection
    */
-  detect_language?: boolean;
+  detect_language?: boolean | string[];
 
   /**
    * @see https://developers.deepgram.com/docs/topic-detection


jkroll-deepgram (approved):

tmqjom:
Hi, just to clarify— is the detect_language feature currently available only for pre-recorded audio? Is it still not supported for live audio streams? 

or are there any ways to apply it on streams. 

https://developers.deepgram.com/docs/language-detection

thanks in advance!

Issue #286, opened 2024-05-10
[Automation] Shellscript Linting


dvonthenen:
## Proposed changes

Shellscript Linting should adhere to known standards. Look into:
- github.com/koalaman/shellcheck

This is already implemented in Go and needs to be ported over:
https://github.com/deepgram/deepgram-go-sdk/pull/208

## Context

> Why is this change important to you? How would you use it? How can it benefit other users?

## Possible Implementation

> Not obligatory, but suggest an idea for implementing addition or change

## Other information

> Anything else we should know? (e.g. detailed explanation, related issues, links for us to have context, eg. stack overflow, codepen, etc)

lukeocodes:
We don't have any shellscripts :/

dvonthenen:
I will open a PR soon that will introduce some, but if we can add some of these things without shell scripts, we can close. 

Reference to this:
https://github.com/deepgram/deepgram-js-sdk/issues/284

Issue #284, opened 2024-05-10
Markdown linting


dvonthenen:
## Proposed changes

Markdown should adhere to known standards. Look into:
ghcr.io/igorshubovych/markdownlint-cli

This is already implemented in Go and needs to be ported over:
https://github.com/deepgram/deepgram-go-sdk/pull/208

## Context

> Why is this change important to you? How would you use it? How can it benefit other users?

## Possible Implementation

> Not obligatory, but suggest an idea for implementing addition or change

## Other information

> Anything else we should know? (e.g. detailed explanation, related issues, links for us to have context, eg. stack overflow, codepen, etc)

PR #322, opened 2024-07-23
Implement markdown and yaml linting, prepare eslint infrastructure

dvonthenen:
This enables some linting/checks for Markdown, GH Action, YAML, and Shell Linting.

This introduces GitHub actions to check those things and allows users to check locally before submitting a PR. Running locally requires `docker`, but the user doesn't necessarily need to install or check manually before submitting a PR. To download all dependencies (ie `make check` = run all lints, checks), they can run `make ensure-deps` to download all dependencies, tools, and etc.

Addresses the following issues: https://github.com/deepgram/deepgram-js-sdk/issues/286 https://github.com/deepgram/deepgram-js-sdk/issues/285 https://github.com/deepgram/deepgram-js-sdk/issues/284

The updates to the Markdown are from the issues raised by the mdlint check.

If you don't feel this adds value to the project, feel free to close this PR and the issues as well.


naomi-lgbt (changes requested):
I love the idea of linting the code, but I'm not sure this approach is best for a JavaScript project?

We don't use shell, or go, or yaml.

And we should be linting the JS/TS with the `eslint` package (and `prettier` if we want to be extra robust). There's also a markdownlint-cli npm package we can use if we want to lint markdown.

lukeocodes (changes requested):
call me old fashioned but i have concerns about this tooling being part of my workflow.

there are so many lighter solutions inside the JS ecosystem that do not require docker to run :/

dvonthenen:
Yup, I get it. It was the easiest lift without investing too much time (as in like 5mins). This isn't required for someone to open a PR as it will check in the GH action. If this is too much for the project to add, you can close this PR.

dvonthenen:
> And we should be linting the JS/TS with the `eslint` package (and `prettier` if we want to be extra robust). There's also a markdownlint-cli npm package we can use if we want to lint markdown.

This doesn't tackle any JS linting.

naomi-lgbt (approved):
I hope you don't mind, but I didn't wanna just discard this entirely.

I pushed a commit to shift the runners into the JS ecosystem, and preserved the rules you configured. 💜

dvonthenen:
Doesn't matter. It's your work and code.

PR #323, opened 2024-08-05
feat: new metadata models endpoints

lukeocodes:
adds new metadata get model endpoints to the JS SDK

### DIFF
--- a/test/manage_rest.test.ts
+++ b/test/manage_rest.test.ts
@@ -228,4 +228,23 @@ describe("making manage REST requests", () => {
     assert.isNotNull(result);
     assert.containsAllDeepKeys(result, ["balance_id"]);
   });
+
+  it("should retrieve a list of models", async () => {
+    const { result, error } = await deepgram.manage.getAllModels(faker.string.uuid());
+
+    assert.isNull(error);
+    assert.isNotNull(result);
+    assert.containsAllDeepKeys(result, ["stt", "tts"]);
+  });
+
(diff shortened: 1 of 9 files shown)


test/manage_rest.test.ts
```diff
@@ -228,4 +228,23 @@ describe("making manage REST requests", () => {
     assert.isNotNull(result);
     assert.containsAllDeepKeys(result, ["balance_id"]);
   });
+
+  it("should retrieve a list of models", async () => {
+    const { result, error } = await deepgram.manage.getAllModels(faker.string.uuid());
+
+    assert.isNull(error);
+    assert.isNotNull(result);
+    // Add your assertions for the response data here
```
naomi-lgbt, line 237: Do we want to test that a specific property is present, like we do in the other tests?

naomi-lgbt (approved):
This looks good! One non-blocking question.

test/manage_rest.test.ts
lukeocodes: shhhhh i hate this logic 

but yes we do :(

jpvajda:
@naomi-lgbt  @lukeocodes  How close are we to getting this merged in?

lukeocodes:
@naomi-lgbt or @dvonthenen if you could review/approve. I have updated the tests to actually check the response body as requested, and commits now squashed

cc @jpvajda

dvonthenen (approved):

PR #324, opened 2024-08-05
fix: process is not defined in browser demos right now

lukeocodes:
When testing the streaming example we found that the BUN changes had introduced a regression for the browser code.

*screams in isomorphic*


naomi-lgbt (approved):
Sounds like the problem is solved by not supporting bun

PR #345, opened 2024-10-28
feat: support voice agent

naomi-lgbt:
PR is just for code review, will manually force push to beta upon approval.

### DIFF
--- a/examples/browser-agent-live/index.html
+++ b/examples/browser-agent-live/index.html
@@ -1,3 +1,7 @@
+<!--
+WARNING: This example is currently non-functional. You may encounter issues
+with browser support during the beta release of the Voice Agent API.
+-->
 <!DOCTYPE html>
 <html>
   <head>


src/packages/AgentClient.ts
```diff
@@ -0,0 +1,21 @@
...
+import { AgentLiveClient } from "./AgentLiveClient";
+
+/**
+ * The `AgentClient` class extends the `AbstractClient` class and provides access to the "agent" namespace.
+ * It exposes one method:
+ *
+ * `live(ttsOptions: SpeakSchema = {}, endpoint = ":version/speak")`: Returns an `AgentLiveClient` instance for interacting with the voice agent API.
+ */
+export class AgentClient extends AbstractClient {
+  public namespace: string = "speak";
```
lukeocodes, line 11:
```suggestion
  public namespace: string = "agent";
```

src/packages/AgentLiveClient.ts
```diff
@@ -0,0 +1,165 @@
...
+export class AgentLiveClient extends AbstractLiveClient {
+  public namespace: string = "agent";
+
+  constructor(options: DeepgramClientOptions, endpoint: string = "/agent") {
+    super(options);
+    /**
+     * According to the docs, this is the correct base URL for the Agent API.
+     * TODO: Make configurable for self-hosted customers.
+     */
+    this.baseUrl = "wss://agent.deepgram.com";
```
lukeocodes, line 15: We should set this as a default in the namespace and let the existing config handler allow overriding it

https://github.com/deepgram/deepgram-js-sdk/blob/7e618010c33334c7603654d0f7f42368e57a6587/src/lib/constants.ts#L47-L49

We'll need another line here like `agent: DEFAULT_GLOBAL_OPTIONS` and define websocket.options.url inside that constant

src/packages/AgentClient.ts
lukeocodes, line 1: I see why we added this file to achieve dot notation for deepgram.agent.live, but I am wondering, as the agent API will never need rest endpoints, whether we forego this file and then we return the AgentLiveClient from `deepgram.agent`

PR #359, opened 2025-02-06
feat: add support for nova 3 and keyterms

naomi-lgbt:


jpvajda (approved):

src/packages/ListenLiveClient.ts
```diff
@@ -34,6 +35,13 @@ export class ListenLiveClient extends AbstractLiveClient {
   ) {
     super(options);
 
+    if (
```
jpvajda, line 38: I was debating if we needed this exception in the SDKs as our API enforcing it.

jpvajda (approved):

src/packages/ListenLiveClient.ts
naomi-lgbt: Discussed in Slack - moot point, we're fine either way as these SDKs will soon be approached differently.

PR #363, opened 2025-02-13
fix: keyterms to keyterm

naomi-lgbt:
## Summary by CodeRabbit

- **Refactor**
	- Renamed the audio processing keyword configuration from a plural to singular form to ensure consistency across live, recorded, and transcription functionalities.
	- Updated the handling of key term inputs to work with the new singular parameter.
	- Deprecated an older session termination method in favor of an improved alternative.

### DIFF
--- a/examples/node-agent-live/index.js
+++ b/examples/node-agent-live/index.js
@@ -28,7 +28,7 @@ const agent = async () => {
       agent: {
         listen: {
           model: "nova-3",
-          keyterms: ["spacewalk"],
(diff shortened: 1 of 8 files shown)


jpvajda (approved):

kdawgwilk:
This actually causes more confusion I think because the old property uses `keywords` which is plural and this does not.  The SDK API I would think should receive `keyterms` plural as the property and then map that to the API `keyterm` query params

PR #381, opened 2025-04-26
fix: instructions -> prompt in updatePrompt to align with v1 spec

jafoobari:
Noticed `updatePrompt` was still using `instructions`. The [the v1 spec](https://github.com/deepgram/VA-API-Spec-v1) expects `prompt`. This updates it to align with the spec.

Ties up the earlier rename from `updateInstructions` → `updatePrompt`.


naomi-lgbt (approved):
Whoops! Thanks for catching this.

PR #383, opened 2025-05-05
feat: release voice agent v1

naomi-lgbt:
BREAKING CHANGE: the interfaces for the voice agent have changed.


naomi-lgbt:
The document change is mostly arbitrary. I need a commit with the very specific message format I've used here so the automation will cut a major release.

jpvajda (approved):

PR #388, opened 2025-05-14
chore: add elevenlabs example

naomi-lgbt:


jpvajda:
@naomi-lgbt What did you want to do with this PR? I assume this might need to change given our approach to TTS providers now being generic, but just confirming if we should try to merge this in or close it out soon.

PR #399, opened 2025-06-13
feat: add Models type for transcription schema

thecuvii:
### Improvement

Add types of model field


lukeocodes:
Our focus is now on releasing V5 which does include model mapping. I'm going to close this PR. Thanks for your contribution!

PR #423, opened 2025-07-21
feat: add support for agent tags

jpvajda:
## TL;DR
Added `tags` field to `AgentLiveSchema` to support request tagging for filtered searching. The field accepts an array of strings at the root level of agent configuration settings.

### Schema Updates
- **File:** `src/lib/types/AgentLiveSchema.ts`
- **Change:** Added `tags?: string[]` field at root level of `AgentLiveSchema` interface
- **Location:** Positioned at top level (not inside `agent` object) per API specification

### Test Coverage
- **File:** `tests/unit/live-client-message-handling.test.ts` 
- **Added:** 5 comprehensive test cases covering:
  - Multiple tags configuration
  - Empty tags array handling
  - Single tag configuration
  - Optional behavior (no tags)
  - Integration with other agent settings

## API Usage

```typescript
const config = {
  audio: { 
    input: { encoding: "linear16", sample_rate: 16000 } 
  },
  tags: ["customer-service", "priority-high", "english"],
  agent: {
    language: "en",
    speak: {
      provider: { type: "deepgram", model: "aura-2-zeus-en" }
    }
  }
};

agent.configure(config);
```

### Schema Structure
```typescript
interface AgentLiveSchema extends Record<string, unknown> {
  experimental?: boolean;
  /**
   * Tags to associate with the request. Can be used for filtered searching.
   */
  tags?: string[];  // ✨ NEW FIELD
  audio: { /* ... */ };
  agent: { /* ... */ };
}
```

### Validation
- ✅ Field is optional (`tags?:`)
- ✅ Accepts array of strings (`string[]`)
- ✅ Works with empty arrays (`tags: []`)
- ✅ Compatible with all existing agent settings
- ✅ TypeScript compilation successful

## Testing Results

```
✅ All Tests Pass (43/43)
  - 38 existing tests (no regressions)
  - 5 new tags-specific tests

✅ Build Success
  - TypeScript compilation ✓
  - Webpack bundling ✓  
  - Prettier formatting ✓

✅ Schema Validation
  - Multiple tags: ✓
  - Single tag: ✓
  - Empty array: ✓
  - Optional (no tags): ✓
  - Combined with other settings: ✓
```

What types of changes does your code introduce to the community JavaScript SDK?
_Put an `x` in the boxes that apply_

- [ ] Bugfix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
[...]

PR #428, opened 2025-08-04
Fix: agent timing issue 

jpvajda:
## PR Summary

Moved this.setupConnection(); to be called after the WebSocket connection is created, not before.

The timing fix was simply ensuring this.setupConnection(); happens at the right moment - after this.conn is established but before the connection opens, so the event handlers are properly attached.

### Files Changed:

- `src/packages/AbstractLiveClient.ts` - Moved this.setupConnection(); to the correct timing position
- `node/agent-live/index.js` example to remove the workaround since the timing was now fixed.

### Testing

- ✅  Testing live with new example approach
- ✅  All Unit Test Pass
- ✅  All e2e Test Pass

### Related

Related to issue found in #423 

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1210966837871377

### DIFF
--- a/examples/node-agent-live/index.js
+++ b/examples/node-agent-live/index.js
@@ -5,39 +5,126 @@ const { createClient, AgentEvents } = require("../../dist/main/index");
 const fetch = require("cross-fetch");
 require("dotenv").config();
 
+console.log("🔑 API Key loaded successfully");
+
 const agent = async () => {
+  // Check for API key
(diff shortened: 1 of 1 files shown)


lukeocodes (approved):
This looks good. Can't imagine why we haven't had more yelling about this though

PR #436, opened 2025-10-22
Refactor HTML for Deepgram Live Transcription

TibinSebastian:
fixed the script and made the page bit user friendly


lukeocodes:
We are replacing these with simpler versions in our upcoming V5 release

PR #480, opened 2026-04-10
fix(listen): warn when keywords is used with nova-3 model

### LINKED ISSUE #474: nova-3 returns HTTP 400 when `keywords` param is used in streaming WebSocket
## Description

When using the Deepgram streaming WebSocket API (`wss://api.deepgram.com/v1/listen`) with `model=nova-3`, including the `keywords` query parameter causes Deepgram to reject the WebSocket upgrade with HTTP 400.

The same request works fine with `model=nova-2`, and using `keyterm` instead of `keywords` works on `nova-3`.

## Steps to Reproduce

```python
import asyncio, websockets

KEY = "your-api-key"

async def test(model, param):
    uri = f"wss://api.deepgram.com/v1/listen?model={model}&language=en-GB&encoding=linear16&sample_rate=24000&channels=1&{param}=Jonty"
    try:
        async with websockets.connect(uri, additional_headers={"Authorization": f"Token {KEY}"}) as ws:
            print(f"{model} + {param}: OK")
            await ws.close()
    except Exception as e:
        print(f"{model} + {param}: FAIL ({e})")

asyncio.run(test("nova-3", "keywords"))   # FAIL - HTTP 400
asyncio.run(test("nova-3", "keyterm"))     # OK
asyncio.run(test("nova-2", "keywords"))    # OK
asyncio.run(test("nova-2", "keyterm"))     # OK
```

## Expected Behaviour

Either:
1. `keywords` should work on `nova-3` (backwards compatible with nova-2), or
2. The API should return a clear error message (not just HTTP 400) indicating that `keywords` is not supported on nova-3 and that `keyterm` should be used instead

## Impact

This is a silent breaking change when upgrading from nova-2 to nova-3. The WebSocket upgrade is rejected with a bare HTTP 400 — no JSON error body is returned, making it very difficult to diagnose. We spent significant time debugging this in production.

The LiveKit agents SDK (`@livekit/agents-plugin-deepgram`) exposes both `keywords` and `keyterm` options and doesn't warn about model compatibility, which compounds the issue.

## Environment

- Deepgram streaming WebSocket API
- Models tested: nova-2 (works), nova-3 (fails)
- Tested with both Python `websockets` and Node.js `ws` packages

peteroyce:
Fixes #474

Adds a runtime `console.warn()` in `WrappedListenV1Client.connect()` when the caller passes both `keywords` and a nova-3 model (`nova-3`, `nova-3-general`, `nova-3-medical`). The Deepgram API silently rejects this combination with an HTTP 400, making it very hard to diagnose.

**Before:** The WebSocket upgrade fails with a bare HTTP 400 — no feedback to the developer.

**After:** A clear warning is printed before the connection attempt:
```
[Deepgram] The 'keywords' parameter is not supported with nova-3 models and will cause
the WebSocket connection to be rejected (HTTP 400). Use 'keyterm' instead.
See: https://developers.deepgram.com/docs/keyterm
```

**Why `CustomClient.ts`:** This file is not auto-generated by Fern and will survive SDK regenerations. The check is placed before the WebSocket is created so developers see the warning immediately.

No breaking changes — this is a warning only, existing behavior is unchanged.

### DIFF
--- a/src/CustomClient.ts
+++ b/src/CustomClient.ts
@@ -629,9 +629,23 @@ class WrappedAgentV1Socket extends AgentV1Socket {
  * consistent behavior across Fern regenerations and allows us to customize
  * connection setup, authentication, and header handling.
  */
+const NOVA3_MODELS = ["nova-3", "nova-3-general", "nova-3-medical"] as const;
+
 class WrappedListenV1Client extends ListenV1Client {
     public async connect(args: Omit<ListenV1Client.ConnectArgs, "Authorization"> & { Authorization?: string }): Promise<ListenV1Socket> {
         const { headers, debug, reconnectAttempts, connectionTimeoutInSeconds, abortSignal } = args;
+
+        // Warn when 'keywords' is used with nova-3 — the API rejects this combination.
+        // 'keywords' is not supported on nova-3; use 'keyterm' instead.
+        const model = (args as Record<string, unknown>).model as string | undefined;
+        const keywords = (args as Record<string, unknown>).keywords;
+        if (keywords != null && model != null && NOVA3_MODELS.some((m) => model === m || model.startsWith(m))) {
+            console.warn(
+                "[Deepgram] The 'keywords' parameter is not supported with nova-3 models and will cause " +
+                "the WebSocket connection to be rejected (HTTP 400). Use 'keyterm' instead. " +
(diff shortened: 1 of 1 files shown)


GregHolmes:
Superseded by #505, which surfaces a catchable error before opening the socket and includes regression coverage.

PR #497, opened 2026-05-06
feat: sdk regeneration 2026-05-06 with backward-compat preserved

GregHolmes:
## Summary

Pulls in the 2026-05-06 Fern regeneration along with the manual patches and backward-compat shims needed to keep the public TypeScript surface stable. No breaking changes for downstream consumers.

### What the regen brings (additive)
- **Audio encoding/container values:** `Mp3`, `Opus`, `Flac`, `Aac` added to `AgentV1Settings.Audio.Output.Encoding`; new `Container` enum (`None`, `Wav`, `Ogg`) replaces the loose `string` typing on `container?` (still includes `string`, so non-breaking).
- **New types:** `AgentV1SettingsAgentContextListenProvider` (with V2 `language_hint` support), `OpenAiThinkProvider`; new fields on `Groq`, `ListProjectMembersV1Response`, `ListenV2ConfigureSuccess`, `SpeakV1Metadata`.
- **Agent-by-ID:** `AgentV1Settings.Agent` is now `{...} | string`, so callers may pass an agent ID directly. Mirrors the Python SDK's `Union[str, AgentV1SettingsAgentContext]`.

### Manual patches re-applied
- `package.json` — restore project devDeps (commitlint, tsx, playwright, vite, terser pinned for the `.npmrc` minimum-release-age policy) and bump `@types/node` to `^20.17.57`.
- `src/api/resources/manage/.../keys/client/Client.ts` — keep the request body optional on `manage.v1.projects.keys.create` so older callers stay source-compatible.
- `src/api/types/CreateKeyV1Request.ts` — preserve the legacy `CreateKeyV1RequestOne` alias.
- `biome.json` — restore `vcs.useIgnoreFile = true` so generated browser artifacts under `examples/` stay out of lint.

### Backward-compat shims (new this regen)
The regen restructured `AgentV1Settings.Agent` from an interface with named sub-types into a union with anonymous shapes, which would have removed `AgentV1Settings.Agent.Context`, `Agent.Listen`, `Agent.Think`, `Agent.Speak`, and `Agent.Context.Messages.Item`.

`AgentV1Settings.ts` is now in `.fernignore` (temporarily-frozen) with a hand-applied namespace block that uses TypeScript declaration merging to re-publish those sub-types as aliases derived from `Exclude<Agent, string>`. Mirrors the pattern used by the Python SDK for the same regen. Regression coverage in `tests/unit/compat-aliases.test.ts` (5 new cases).

The parallel `AgentV1SettingsAgentListenProvider` → `AgentV1SettingsAgentContextListenProvider` rename gets a one-line type alias (with namespace merge for `.V1`/`.V2`) so pinned-name callers (`const p: AgentV1SettingsAgentListenProvider = ...`) keep compiling without a cast. Purely additive at the type level — the new V2 only adds an optional `language_hint`. File added to `.fernignore`; `Equals<X, Y>` identity assertions in `tests/unit/compat-aliases.test.ts`.

### Repository hygiene
- One-time `biome format --write` pass over 91 hand-maintained files that were authored with 2-space indent before the project's `biome.json` standardised on 4-space. Pure whitespace, no behaviour change. `make lint` is now clean.

## Test plan
- [x] `make build` — TypeScript and ESM build pass
- [x] `make test` — 533 unit + 132 wire tests pass (compat-aliases now covers Agent namespace, listen-provider alias, and `keys.create` no-arg)
- [x] `make test-esm` — ESM build validation pass
- [x] `make lint` — clean (down from 91 format errors)
- [x] Verify `AgentV1Settings.Agent.Context`, `Agent.Listen`, `Agent.Context.Messages.Item` etc. still resolve via the regression test
- [x] Confirm `AgentV1SettingsAgentListenProvider` is type-identical to `AgentV1SettingsAgentContextListenProvider` via the alias


lukeocodes (approved):
mostly clean regen. all the back-compat shims are intact on the head branch and `tests/unit/compat-aliases.test.ts` locks the alias names with `Equals<X, Y>` identity checks that fail at compile time if a future regen drifts them. nothing here breaks an existing caller.

what the shims achieve:

- `AgentV1Settings.Agent` would have collapsed from an interface with named sub-types into `{...} | string`, deleting `Agent.Context`, `Agent.Listen`, `Agent.Context.Messages.Item`, `Agent.Think`, `Agent.Speak`. the `.fernignore` swap restores the interface form so direct reads like `settings.agent.context?.messages` keep type-checking in strict mode without `typeof` narrowing.
- `AgentV1SettingsAgentListenProvider` was renamed to `AgentV1SettingsAgentContextListenProvider`. one-line type alias plus namespace declaration-merge republishes `.V1` and `.V2`, both names co-exported from the resource barrel. legacy V2 literals without `language_hint` still compile (covered by `tests/unit/compat-aliases.test.ts:194-203`).
- `keys.create(project_id)` no-body call site stays valid via the existing `request?` optionality patch on `KeysClient`. compat-aliases test exercises both the no-body and with-body forms.

two things worth flagging, neither blocking:

1. `Audio.Output.container` tightens from `string | undefined` to `Container | undefined` where `Container = "none" | "wav" | "ogg" | string`. the `| string` fallback keeps runtime safe and old callers passing arbitrary strings still type-check, but no test pins that fallback. one-line literal-assignment in `compat-aliases.test.ts` would lock it:

   ```ts
   const _looseContainer: AgentV1Settings.Audio.Output = { container: "flac" };
   ```

2. `AgentV1Settings.AgentReference = Agent | string` is the right escape hatch for the new agent-by-ID flow, but `AgentV1Settings.agent` is still typed `Agent`, not `AgentReference`. so:

   ```ts
   const settings: AgentV1Settings = { type: "Settings", audio: {...}, agent: "agent_123" }; // TYPE ERROR
   ```

   users only discover this when they try. the python sdk exposes the equivalent as `Union[str, AgentV1SettingsAgentContext]` at the parent level. either add a docstring on `AgentReference` showing the cast pattern, or queue typing the parent field as `AgentReference` for v6.

everything else is purely additive: `Audio.Output.Encoding` (+ `Mp3`/`Opus`/`Flac`/`Aac`), `Audio.Output.Container` enum, `AgentReference` alias, `AgentV1ConversationText.languages_hinted`/`languages`, `ListenV2ConfigureSuccess.language_hints`, `SpeakV1Metadata.additional_model_uuids`, `Groq.reasoning_mode` + `ReasoningMode` enum, new `OpenAiThinkProvider`, `ListProjectMembersV1Response.Members.Item` (+ `scopes`/`first_name`/`last_name`, wire-tested in `tests/wire/manage/v1/projects/members.test.ts:17-25`).

`src/CustomClient.ts` is 100% biome reformatting despite the 82+60- diff. 91 hand-maintained files reformatted with `biome format --write` over 2-space → 4-space indent, no behaviour change.

ship it.

PR #505, opened 2026-06-04
fix(listen): reject Nova-3 keywords before connecting

### LINKED ISSUE #470: WebSocket connection fails with keywords parameter (Nova-3, 1006)
## Description

WebSocket connections to `wss://api.deepgram.com/v1/listen` fail with close code 1006 when the `keywords` parameter is included in the URL. Without `keywords`, connections work fine every time.

## Reproduction

**Works (connects successfully):**
```js
const ws = new WebSocket(
  'wss://api.deepgram.com/v1/listen?model=nova-3&language=en-US&encoding=linear16&sample_rate=16000&channels=1',
  ['token', '<temporary_token>']
);
// → onopen fires, connection established
```

**Fails (1006 every time):**
```js
const ws = new WebSocket(
  'wss://api.deepgram.com/v1/listen?model=nova-3&language=en-US&encoding=linear16&sample_rate=16000&channels=1&keywords=TCP%3A2&keywords=UDP%3A2',
  ['token', '<temporary_token>']
);
// → onerror fires, onclose with code 1006
```

Even with just 2 simple keywords (`TCP:2` and `UDP:2`), the connection is rejected.

## Environment

- **Model:** Nova-3
- **Auth:** Temporary API keys via `POST /v1/keys` (project-scoped)
- **Client:** Browser WebSocket (Chrome 134, macOS)
- **Connection method:** `new WebSocket(url, ['token', tempKey])`
- **Started:** March 19, 2026 — same code was working fine the day before with no changes on our end

## Expected behavior

WebSocket should connect successfully with `keywords` parameter, as documented in the [Deepgram API docs](https://developers.deepgram.com/docs/keywords).

## Additional context

We're building [BufferMind](https://buffermind.com), a real-time lecture transcription app. We use keyword boosting to improve recognition of academic terminology (course-specific terms). This was working reliably until March 19.

Contacted Trevor Christensen (Product Consultant) who suggested posting here.

### LINKED ISSUE #474: nova-3 returns HTTP 400 when `keywords` param is used in streaming WebSocket

valeriudev:
## Summary

Nova-3 uses `keyterm` prompting instead of the legacy `keywords` path. The API intentionally rejects non-empty `keywords` before upgrading a Listen V1 WebSocket, but browsers cannot read the rejected handshake body and developers receive no actionable error.

This adds a pre-connect guard in `WrappedListenV1Client.connect()` that throws the generated `BadRequestError` with the same status, code, and guidance as the API. Nova-2 and older models continue to support `keywords`, while Nova-3 continues to support `keyterm`.

Fixes #474 and #470. Supersedes the now-closed #480.

## Behavior

| Case | Result |
| --- | --- |
| Nova-3 family + non-empty `keywords` | `BadRequestError` with status 400 before dialing |
| Nova-3 + `keyterm` | Connection object created normally |
| Nova-2 + `keywords` | Connection object created normally |
| Empty `keywords` | Connection object created normally |
| Similar names such as `nova-30` | Not treated as Nova-3 |

## Implementation

- Keeps the guard in hand-maintained `src/CustomClient.ts`, so no generated source is patched.
- Uses exact Nova-3 family matching: `nova-3` or the `nova-3-` prefix.
- Adds public-client coverage to the existing frozen `tests/unit/custom-client.test.ts`; no new test file or `.fernignore` entry is added.

## Verification

- [x] Live production handshake: `nova-3 + keywords` returns HTTP 400.
- [x] Live production handshake: `nova-3 + keyterm` opens successfully.
- [x] Live production handshake: `nova-2 + keywords` opens successfully.
- [x] Build passes for CJS and ESM.
- [x] Compatibility typecheck passes.
- [x] Unit tests pass: 644/644.
- [x] Wire tests pass: 142/142.
- [x] Focused custom-client tests pass: 38/38.
- [x] `git diff --check` passes.


valeriudev:
Heads-up for triage: this overlaps **#480** (@peteroyce), which also fixes #474. The difference is #480 logs a `console.warn` then still connects (so the socket dies silently), whereas this throws the same `BadRequestError` the REST path already does — catchable via `instanceof` on both transports — plus a regression test. Credit to @peteroyce for the original diagnosis; happy to close in favour of #480 if the lighter-touch warning is preferred.

GregHolmes (approved):

PR #509, opened 2026-06-16
feat(regen): flux word timings, diarize_model, profanity_filter — fully backward-compatible

GregHolmes:
## SDK Regeneration — 2026-06-16

Fern generator output reconciled against hand-maintained patches. **Fully backward-compatible — no breaking changes vs `main`.** Ships as a minor release.

### Features (additive, Fern-generated)
- **`diarize_model` + `DiarizeModel`** on listen v1 (the legacy `diarize` is deprecated, not removed).
- **`profanity_filter` + `ListenV2ProfanityFilter`** on listen v2.
- **Flux word-level timings** — optional `start`/`end` on `ListenV2TurnInfo` words.
- New `isAuthProvider` helper + `auth` option on the base client.

### Backward-compatibility shims (kept this non-breaking)
- **`ListenV2CloseStream.Type`** — the regen dropped the `Type` namespace. A CloseStream message's `type` is only ever `"CloseStream"`; `Finalize`/`KeepAlive` were copied in error from v1's control-message enum. Rather than remove those members (a break for any `Type.Finalize` reference), the `Type` namespace is restored to match `main` with `Finalize`/`KeepAlive` marked **`@deprecated`** (slated for next-major removal), keeping `| string` openness. Existing references compile unchanged. Frozen.
- **`language_hint` → `language_hints`** (`DeepgramListenProviderV2`) — `language_hint` is re-added as a deprecated optional field so existing call sites keep compiling (the singular field was never honored by the API; prefer `language_hints`). The `DeepgramListenProviderV2.LanguageHint` namespace type the regen removed is recreated via declaration merge. Frozen.

### Version handling — single-sourced
- **`src/version.ts`** is the single source of truth: carries the `// x-release-please-version` marker and is the only generic `extra-file` release-please updates. Frozen to preserve the marker; kept at `5.4.0` (release-please bumps on merge).
- **`src/BaseClient.ts`** derives the version via `import { SDK_VERSION }` instead of a hardcoded string, so the User-Agent can't drift.

### Test / typecheck hardening
- **`tests/unit/compat-aliases.test.ts`** — fixed imports to use the `Deepgram.agent.*` namespace (the `AgentV1Settings*` types aren't on the top-level barrel), so the file now type-checks. The `Equals<>` assertion errors were artifacts of the broken imports, not shim drift — every assertion holds.
- **Added a real typecheck gate** — `tsconfig.typecheck.json` + `make typecheck-tests`, wired into `make lint` and CI, so the compat-alias / regen-constraint assertions actually gate (previously `tsc` only covered `src/` and vitest strips types).
- **`tests/unit/regen-constraints.test.ts`** (new, 7 tests) — covers the CloseStream `Type` shim (incl. the deprecated `Finalize`/`KeepAlive`), `language_hints` round-trip + deprecated `language_hint`, `ListenV2ProfanityFilter`, and optional word `start`/`end`.

### Other re-applied frozen patches
`package.json` (devDependency additions + pins), `biome.json` (`useIgnoreFile`), `src/Client.ts` (`client.fetch()` defaults to `base`, not `agentRest`), `CreateKeyV1Request.ts` (`CreateKeyV1RequestOne` alias), manage-keys `Client.ts` (optional `request?` body), `AgentV1Settings.ts` (`Agent` namespace interface form + opt-in `AgentReference`).

### Generator-owned now (no longer a frozen patch)
- **`AgentV1SettingsAgentListenProvider.ts`** — Fern now emits a type structurally equal to `AgentV1SettingsAgentContextListenProvider` (verified by the `Equals<…>` assertion in `compat-aliases.test.ts`), so it's modified by the generator rather than hand-frozen. Fern owns the file again.

### Verification
- `make build` — ✅
- `make test` — ✅ 557 unit + 132 wire
- `make lint` — ✅ (biome lint + format + the new `typecheck-tests` gate)
- `ListenV2CloseStream.Type` members match `main` (no public-surface removal); word `start`/`end` optional; `language_hint` retained — no breaking changes vs `main`.

`pnpm-lock.yaml` rebuilt via `pnpm install` after the `package.json` patch. All `.bak` files removed; `.fernignore` + `AGENTS.md` updated.


dg-coreylweathers (approved):
Approving as code owner. Backward-compatible SDK regen; review blockers from REVIEW-js-509.md addressed in latest commits (typecheck gate, regen-constraint tests, Finalize/KeepAlive kept deprecated). Shipping as minor 5.5.0.

PR #511, opened 2026-06-26
docs(tests): adding code coverage processes on CI.yml workflow

mgama-deepgram:
## Proposed changes

- Adding code coverage processes on the CI.yml workflow on the test job
- The test job in the CI.yml workflow will now post the calculated code coverage as a PR comment

## Types of changes

What types of changes does your code introduce to the community JavaScript SDK?
_Put an `x` in the boxes that apply_

- [ ] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Documentation update or tests (if none of the other choices apply)

## Checklist

_Put an `x` in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before merging your code._

- [x] I have read the [CONTRIBUTING](../CONTRIBUTING.md) doc
- [x] I have lint'ed all of my code using repo standards
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have added necessary documentation (if appropriate)

PR #514, opened 2026-06-30
test: increase unit test coverage including Speak V2 Flux TTS

mgama-deepgram:
## Proposed changes

- Adding code coverage processes on the CI.yml workflow on the test job
- The test job in the CI.yml workflow will now post the calculated code coverage as a PR comment
- Increases code coverage (aided by analysis via Claude): **line coverage 79% → 97%, branch coverage 67% → 94%** (full unit + wire suite, measured over the whole `src` tree)
- Replaces original PR to only calculate test coverage https://github.com/deepgram/deepgram-js-sdk/pull/511

### Summary

Adds a comprehensive hand-written unit test suite across the SDK's hand-maintainable surface (core utilities, transport/websocket layer, custom client, generated REST/WS clients, and Speak V2), wires **V8 code coverage into CI** with an automatic PR comment and a threshold gate, and repairs a lockfile broken by an earlier merge. All new tests are frozen in `.fernignore` so they survive the next Fern regeneration.

**18 new test files**, `1235` unit tests passing (`1379` across the full unit + wire suite, 90 files).

### Coverage

All figures below are measured the same way — `make test-coverage` (unit + wire in one vitest invocation) over the whole `src` tree with no omit list — so the before/after delta reflects what the new tests actually bought.

| | Statements | Branches | Functions | Lines |
|---|---|---|---|---|
| `main` (72 files / 788 tests) | 78.99% | 66.76% | 80.99% | 79.04% |
| This branch (90 files / 1379 tests) | **97.30%** | **94.19%** | **96.98%** | **97.37%** |

For reference, the unit project alone now runs 1235 tests and reaches 94.77% lines; the full-suite numbers above are the ones CI publishes, since the 30 pre-existing wire test files already cover a meaningful share of `src`.

### What changed

**CI (`.github/workflows/ci.yml`)**
- Runs on `pull_request`, with `push` narrowed to `main` — previously `on: [push, pull_request]` ran the entire workflow twice for every push to a branch with an open PR.
- Test step now runs `make test-coverage` (unit + wire under V8 coverage).
- Generates a coverage summary (CodeCoverageSummary), posts/updates a **sticky PR comment**, and writes it to the job summary. Grants the `test` job `pull-requests: write`.
- `examples` and `test-browser` are skipped on fork PRs: GitHub withholds repository secrets there, so both jobs would hard-fail on a contributor's PR for reasons they cannot fix. The coverage comment step is gated the same way and marked `continue-on-error`.
- Both third-party actions are pinned to commit SHAs rather than mutable tags, since the `test` job now holds `pull-requests: write`.

**Tooling**
- **`Makefile`** — new `test-coverage` target: runs unit + wire in a single vitest invocation so coverage merges into one cobertura + text report (CI reads `coverage/cobertura-coverage.xml`); the wire-import fixup is reverted even on failure. Listed in `make help`.
- **`vitest.config.mts`** — normalized project `root`/`include` paths and added a `coverage` block (v8, `src/**/*.ts`, `./coverage`) with **thresholds** (95 statements / 92 branches / 95 functions / 95 lines) so coverage is a floor rather than just a dashboard.
- **`.gitignore`** — ignore `coverage/`, which also keeps the generated report out of biome's lint set.
- **`.fernignore`** — froze all new hand-written test files (and the `tests/unit/core` dir) so a regen won't delete them; entries sorted.
- **`AGENTS.md`** — documented `@vitest/coverage-v8` in the `package.json` regen patch list so it is re-applied after the next regeneration.

**New test coverage** (new files unless noted)
- **Core:** `core/polyfills`, `core/reconnecting-websocket`, `core/runtime-detection`, `core/utils`, `core-branches`, `misc-branches`, `fetcher-branches`.
- **Transport / websockets:** `websocket-native`, `websocket-sockets` (generated Socket classes), `generated-ws-clients` (generated `connect()` query serialization).
- **Custom client:** `custom-client-internals`, `custom-client-browser`, `custom-client-fallback`.
- **REST / misc:** `rest-clients`, `barrels`, `auth/HeaderAuthProvider`, `file/multipart`.

**Lockfile repair (`package.json`, `pnpm-lock.yaml`)**
- Fixed a `pnpm-lock.yaml` broken by a bad merge (dependency graph referenced `cosmiconfig@9.0.2` while only `9.0.1` was defined → `ERR_PNPM_LOCKFILE_MISSING_DEPENDENCY` in CI).
- Rebased the lockfile on main's clean version and exact-pinned both `vitest` and `@vitest/coverage-v8` to `4.1.11`. They peer-depend on the same minor and `.npmrc` sets `strict-peer-dependencies=true`, so a caret on either lets them drift apart into a failed install. `pnpm install --frozen-lockfile` now passes.

---

### 🎙️ Speak V2 (Flux TTS) — coverage changes

All Speak V2 changes in this PR, grouped for review:

**New: `tests/unit/speak-v2-coverage.test.ts`** (mirrors the Python SDK's `test_speak_v2_coverage.py`)
Raises coverage of `src/api/resources/speak/resources/v2/**` from **78% → 100% lines / 56% → 100% branches** (measured full-suite, matching the table above):

| File | Before | After |
|---|---|---|
| `client/Client.ts` | 28.6% | **100%** |
| `client/Socket.ts` | 82.0% | **100%** |
| `resources/audio/client/Client.ts` | 64.3% | **100%** |

Covers the gaps left by the existing split tests:
- **`Client.ts` `connect()`** (absent from `generated-ws-clients.test.ts`) — query-param serialization (string vs `toJson` branches), absent-param omission, `queryParams`/headers/options merge, and the `audio` lazy getter. Uses an aborted signal so no real socket opens.
- **`Socket.ts`** (absent from the `describe.each` in `websocket-sockets.test.ts`) — event forwarding, `readyState`, send guards + open path, `connect()`/`close()` teardown, all three `waitForOpen()` cases, `sendBinary`, and the `Socket is not connected` guard.
- **`audio/client/Client.ts`** (`speak-v2-batch.test.ts` only drove the 2xx path) — the `400 → BadRequestError`, non-400 `→ DeepgramError`, and transport-failure `→ handleNonStatusCodeError` branches, plus an all-optional-params success case. The `?? Production` environment fallback (`audio/Client.ts:79`) is covered without touching the network by injecting a capturing fetcher, so the subtree reaches 100% with no live host calls.

**Modified: `tests/unit/regen-constraints.test.ts`**
Added two tests pinning the **complete** member set of `SpeakV2Encoding` and `SpeakV2SampleRate`. The pre-existing test only asserted one member per enum — a regen dropping/renaming a value (e.g. removing the `44100` sample rate) would have kept coverage at 100% but slipped through. Both also assert the open-string union still holds. (The other `SpeakV2*` types under `src/api/types` are pure type aliases with no runtime code, so they're already effectively fully covered.)

**`.fernignore`** — froze `tests/unit/speak-v2-coverage.test.ts` (`regen-constraints.test.ts` was already frozen).

## Types of changes

What types of changes does your code introduce to the community JavaScript SDK?
_Put an `x` in the boxes that apply_

- [ ] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Documentation update or tests (if none of the other choices apply)

## Checklist

_Put an `x` in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before merging your code._

- [x] I have read the [CONTRIBUTING](../CONTRIBUTING.md) doc
- [x] I have lint'ed all of my code using repo standards
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have added necessary documentation (if appropriate)


mgama-deepgram:
@GregHolmes this is ready for another round of review. I added more test coverage for Speak V2 (Flux TTS). Thanks in advance

tests/unit/speak-v2-coverage.test.ts
```diff
@@ -0,0 +1,476 @@
...
+        });
+        socket.on("error", (e) => {
+            seen.error = e;
+        });
+
+        fake.emit("open", { type: "open" });
+        expect(seen.open).toBe(true);
+
+        fake.emit("message", { data: '{"type":"Flushed","sequence_id":1}' });
+        expect(seen.message).toEqual({ type: "Flushed", sequence_id: 1 });
```
jherl-dg: The actual messages uses `Flushed.speech_id` not `Flushed.sequence_id`

jherl-dg (commented):
1 small thing. Other than that the /v2/speak coverage here and in python look accurate! Defer to corey for the official stamp here tho

tests/unit/speak-v2-coverage.test.ts
mgama-deepgram: Thanks for calling this out @jherl-dg I just updated the test to match the correct property `Flushed.speech_id`

mgama-deepgram:
@dg-coreylweathers @GregHolmes these changes are ready to be re-review. I have addressed the original changes requested from the first review. Thanks in advance

mgama-deepgram:
@GregHolmes @dg-coreylweathers this PR is ready for a final re-review. Thanks in advance for getting approved and merged the python-sdk PR counterpart

GregHolmes (changes requested):
# Review: deepgram-js-sdk #514, test: increase unit test coverage including Speak V2 Flux TTS
Classification: code
Verdict: request-changes

## Intent

- Raise meaningful unit and whole-suite coverage across the SDK without changing runtime behavior.
- Enforce a coverage floor in CI and publish an updated coverage summary on same-repository pull requests.
- Avoid duplicate branch/PR workflow runs and avoid secret-dependent jobs on fork pull requests.
- Preserve every new hand-written test through Fern regeneration.
- This is a permanent, human-authored test/CI change replacing PR #511; no external launch gate is identified.

## Blocking

### [B2] Pull-request code runs with an avoidable write-capable token

**Summary:** `.github/workflows/ci.yml:73-120` grants the entire `test` job `pull-requests: write`, then checks out PR code with persisted credentials, installs its dependencies, and runs its tests before generating the comment. The SHA-pinned CodeCoverageSummary action also delegates execution to the mutable image tag `ghcr.io/irongut/codecoveragesummary:v1.3.0`, so the comment's claim that pinning prevents replacement is incomplete. A compromised dependency, same-repository PR, or replaced container image can use the job token to mutate pull-request comments.

**Expected:** Untrusted PR code and mutable third-party executables run with read-only permissions; only the minimal reporting step receives `pull-requests: write`.

**Observed:** PR-controlled install/test execution and a mutable Docker image share the same write-capable job and checkout credential as the sticky-comment step.

**Recommended fix:** Keep coverage generation in a read-only test job with `actions/checkout` configured as `persist-credentials: false`, upload the generated Markdown as an artifact, and move the SHA-pinned sticky-comment action into a separate same-repository-PR-only job with `pull-requests: write` that never checks out or executes PR code.

## Should-fix

### [S1] Several tests pass without exercising the behavior named in their titles

**Summary:** `tests/unit/websocket-sockets.test.ts:188-207` converts either promise rejection or resolution into a defined value, so its rejection test passes both ways; its “invalid JSON” test sends valid JSON. `tests/unit/custom-client-internals.test.ts:379-399` deletes `ping` from an instance even though the method lives on `FakeTransport.prototype`, so the alleged no-ping branch still invokes a real `ping` method. These tests increase execution counts without guarding the stated contracts.

**Expected:** Each test fails when the named rejection, malformed-message, or missing-ping behavior regresses.

**Observed:** The assertions remain green when `waitForOpen()` resolves incorrectly, no malformed JSON is sent, and the transport still exposes `ping()`.

**Recommended fix:** Assert the original promise with `await expect(promise).rejects.toEqual(...)`; either send genuinely malformed input and assert supported behavior or rename/remove that test; construct a transport object that truly omits `ping` instead of deleting an inherited method.

### [S2] Speak V2 coverage freezes incorrect repeatable-tag serialization

**Summary:** `tests/unit/speak-v2-coverage.test.ts:86-98` expects `tag: ["t1", "t2"]` to become one JSON string. `src/api/types/SpeakV2Tag.ts:3-7` documents `tag` as repeatable, whose wire form is repeated query parameters. The test inspects a private `_queryParameters` object instead of the encoded upgrade URL, so it codifies generated implementation behavior rather than the public wire contract.

**Expected:** Repeatable tags reach the wire as `tag=t1&tag=t2`, and a regression test observes the final encoded URL through a supported client path.

**Observed:** The frozen test requires the single value `tag=["t1","t2"]` and would reject a future generator fix that emits the documented repeatable form.

**Recommended fix:** Remove the JSON-string expectation, add a URL-capturing wire/public-client assertion for repeated tags, and treat any exported generated-client mismatch as a separate compatibility bug rather than preserving it for coverage.

### [S3] Coverage comments use different thresholds and disappear on failures

**Summary:** `vitest.config.mts:42-56` enforces 95% statements/functions/lines and 92% branches, but `.github/workflows/ci.yml:107-138` omits CodeCoverageSummary's `thresholds` input, so its badges and health marks use the action default of 50/75. The summary action also has the implicit `success()` condition, so a test or threshold failure skips report generation even when Cobertura XML exists; the always-running final step then reports only “No coverage report generated.”

**Expected:** The PR comment visually agrees with the enforced floor and provides coverage diagnostics when the floor fails.

**Observed:** Packages can receive a green health mark at 75% while CI enforces 95%, and failing coverage produces no actionable summary.

**Recommended fix:** Pass an aligned line threshold such as `thresholds: "95 95"` or remove the ambiguous indicators, run summary generation under `if: always()` when the XML exists, and allow the sticky/job summary path to publish the report even when the coverage command exits nonzero.

### [S4] Speak V2 unit tests depend on fixed real TCP ports

**Summary:** `tests/unit/speak-v2-coverage.test.ts:306-405` binds localhost port 39997 without an error rejection path and assumes port 1 is closed to simulate transport failure. Either port can be occupied or behave differently in a developer/CI environment, turning a deterministic unit suite into a collision, timeout, or false result.

**Expected:** Unit tests use injected fetch implementations or OS-assigned ephemeral ports and fail promptly with controlled errors.

**Observed:** Test success depends on two machine-level port assumptions, with the failed-listen path able to surface as an unhandled error or hook timeout.

**Recommended fix:** Use the existing capturing/injected fetcher for status and thrown-network-error branches. If an HTTP server remains necessary, listen on port `0`, read the assigned port, and reject the setup promise on `error`.

### [S5] Coverage cleanup failures are masked

**Summary:** `Makefile:334-341` saves the Vitest status, runs `revert-wire-test-imports.js`, then exits with only the saved Vitest status. If tests pass but the revert script fails, the target exits zero while leaving generated wire imports modified; signals or cancellation bypass cleanup entirely because there is no shell trap.

**Expected:** The target restores wire imports on normal failure and interruption, and either a test failure or cleanup failure makes the target fail.

**Observed:** A failed cleanup after passing tests is reported as success, and interrupted runs can leave the worktree rewritten.

**Recommended fix:** Install an `EXIT INT TERM` trap before the test command, preserve both test and cleanup statuses, and return nonzero when either operation fails.

## Nits

None.

## Verified (evidence)

- PASS: Former B1 is resolved. PR #514 is rebased directly onto `bc6f9c4`, GitHub reports it conflict-free and mergeable, and current v5.9.0 freeze entries, examples, compatibility patches, and WebSocket proxy tests are retained.
- PASS: `pnpm install --frozen-lockfile --no-optional=false` succeeds with `ws` 8.21.3 and exact-matched `vitest` / `@vitest/coverage-v8` 4.1.11.
- PASS: `make test-coverage` runs 1,379 tests across 90 files and reports 97.30% statements, 94.19% branches, 96.98% functions, and 97.37% lines.
- PASS: Current `main`, measured with the same project roots/include set, runs 788 tests across 72 files and reports 78.99% statements, 66.76% branches, 80.99% functions, and 79.04% lines.
- PASS: Coverage cleanup restored all 22 rewritten wire-test imports in the ordinary success path.
- PASS: `make build`, ESM validation, `make test`, `make test-esm`, `make lint`, and `make typecheck-tests` pass on the rebased head.
- PASS: `actionlint` 1.7.7 in Docker reports no workflow syntax errors.
- PASS: all new test files are covered by explicit `.fernignore` entries or the `tests/unit/core` directory entry.
- PASS: fork PRs avoid `pull_request_target`, and the secret-dependent browser/examples jobs are skipped for ordinary forks.
- PASS: the prior Speak V2 review thread was corrected and resolved.
- PASS: The PR body and coverage comment now report the rebased test counts, dependency versions, and before/after coverage metrics.
- PASS: GitHub compile, coverage/test, examples, CodeQL, title, and browser checks all pass at current head `32b23c9`. The prior rebased head had one transient Text Intelligence browser failure that passed unchanged on rerun.
- FAIL: several tests and the coverage-report UI do not enforce or communicate the behavior claimed by their names.

## Needs human

- Greg should decide whether a 95% gate over the entire Fern-generated `src` tree is the intended long-term maintenance policy. Recommendation: keep it only if every regeneration is expected to add/update tests before merge; otherwise gate hand-maintained code separately and report generated-code coverage without using it as a hard global floor.
- The repository already exposes live Deepgram secrets to code on same-repository feature branches in examples/browser CI. PR #514 does not create that trust boundary, but its workflow rewrite preserves it. Recommendation: track protected-environment or post-merge-only live tests as a separate CI-hardening task.

## Developer-facing messaging

Coverage reporting is useful only when its green/red language matches the actual merge gate and its tests fail on behavioral regressions. After the rebase, token isolation, assertion fixes, and aligned reporting, this can provide a trustworthy ratchet instead of a high percentage that hides false-positive tests or blocks valid generator evolution for the wrong reason.

mgama-deepgram:
Thanks @GregHolmes for your input. I have addressed the concerns on my latest commit.

GregHolmes (approved):

Issue #89, opened 2022-11-28
Importing @deepgram/sdk/browser is throwing: SyntaxError: Unexpected token 'export'


Limeslices:
## What is the current behavior?

>When trying to do the following
```
import { Deepgram } from '@deepgram/sdk/browser'
```

I run into this error
```
SyntaxError: Unexpected token 'export'
```

This is running on a NextJs localhost environment.

IvanCoronado:
I have the same error. The browser version is not exported...

It looks like the doc is outdated.

briancbarrow:
@Limeslices @IvanCoronado what version of the SDK were you using? Sorry for the late response. I had an example of this syntax working in v1.11.0 and still in v1.18.0. 

Let me know if you are still running into issues.

lukeocodes:
Closing due to the age and no followups. Please use the latest version and raise a new issue if this continues to be a problem.

PR #526, opened 2026-07-31
feat(regen): listen v2 ForceEndTurn + redact, diarization info, Flux TTS voices

GregHolmes:
## Summary

Fern SDK regeneration for 2026-07-31, with manual patches reconciled against the new generator output.

Generator `fernapi/fern-typescript-sdk` 3.72.4, CLI 5.89.0, API definition pinned at `9e641c9` (`originGitCommitIsDirty: true`).

## New API surface

- **Listen v2:** `ForceEndTurn` client message (`socket.sendForceEndTurn()`), `redact` connect param (`ListenV2Redact`: `numbers` / `aggressive_numbers`), `TurnInfo.trigger` (open enum: `model` / `manual` / `timeout`), `eot_threshold` range extended to 1.0
- **Listen v1:** `diarize_info` on response metadata (with `DiarizeInfo.Arch` v1/v2), `speaker` + `speaker_confidence` on channel word items
- **Speak:** 12 new `flux-*` Flux TTS voices; `Deepgram.version` widened `"v1"` → `string`. ⚠️ Flux TTS is **Early Access** — surface and voice catalog may change before GA
- **Agent:** new `AgentV1UpdateListenListenProvider` union (explicit v1/v2 listen provider selection mid-session)

All of the above is additive. Verified via export/enum sweeps: **no removed exports, no removed enum members.**

## Back-compat

One breaking change was found in review and shimmed (see `a6e3c79`):

- **`AgentV1UpdateListen.Listen.provider`** was repointed from `DeepgramListenProviderV2` to the new `AgentV1UpdateListenListenProvider` union, whose `V1`/`V2` variants both redeclare `version` as **required**. Since `DeepgramListenProviderV2.version` is optional, callers passing `{ type: "deepgram", model: "flux-general-en" }` stopped compiling (TS2322) on a published send path (`agent.v1` → `sendUpdateListen`). The union's `V1` arm also declares `model?`, widening reads of `provider.model` from `string` to `string | undefined`.
- **Fix:** replace the field with a single merged shape — `Omit<DeepgramListenProviderV2, "version"> & { version?; language?; smart_format? }` — that never references the generated union, restoring the pre-regen contract for both construction and reads while keeping `model` required. (A union can't work: property access on a union requires the field on *every* arm, so V2-only reads like `eot_threshold`/`language_hints` would break with TS2339 on the `V1` arm.) Compile-compat only — serialized payload unchanged. File frozen in `.fernignore`, documented in `AGENTS.md`, regression-covered in `tests/unit/regen-constraints.test.ts` (verified: reverting the shim fails `make typecheck-tests` with the exact TS2322).
- The upstream spec *relaxed* this field (its docs dropped "the provider identity (type, version, model) is required and must match the current session"), so the required discriminant looks like a Fern modeling artifact — worth reporting upstream so the shim can be dropped.

### Re-applied (still needed — kept frozen)
`biome.json` (vcs/useIgnoreFile) · `package.json` (extra devDeps + `@types/node`) · `AgentV1Settings.ts` (interface form of `Agent` + `AgentReference`) · `AgentV1SettingsAgentContextListenProvider.ts` (nested `LanguageHint`) · `ListenV2CloseStream.ts` (`Type` namespace) · `keys/client/Client.ts` (optional `request?`) · `CreateKeyV1Request.ts` (`CreateKeyV1RequestOne`) · `DeepgramListenProviderV2.ts` (deprecated `language_hint` + `LanguageHint`) · `BaseClient.ts` (`SDK_VERSION` import) · `Client.ts` (`base` not `agentRest`) · `version.ts` (release-please marker)

All 11 verified byte-identical to `main` (zero drift), and each generator change to them was confirmed to be *only* the removal of our patch — nothing legitimate discarded.

### Newly frozen
- `AgentV1UpdateListen.ts` — the provider shim above.

### Dropped (generator caught up — Fern now owns it)
- `AgentV1LatencyReport.ts` — `stt_latency` returned to the spec; Fern emits `stt_latency?: number` natively with an identical signature, so the hand-added shim was redundant. **Not a breaking change.** Regression test retained as a forward guard.

### Version
Fern bumped 5.7.0 → 5.7.1; rolled back to **5.7.0** so `src/version.ts` (release-please single source of truth, marker intact) stays authoritative across `version.ts` / `package.json` / `.fern/metadata.json`.

## Cross-SDK alignment (python #750, java #89)

Both sibling regen PRs for the same 2026-07-31 spec were checked; JS matches their calls.

| Item | Python #750 | Java #89 | JS (this PR) |
|---|---|---|---|
| `AgentV1UpdateListen` provider required `version` | shimmed — `model_validator` coerces a legacy `DeepgramListenProviderV1/V2` | shimmed — `getProvider()` still returns `DeepgramListenProviderV2`; union via `getProviderValue()` | shimmed — merged shape preserving all prior reads/writes |
| `stt_latency` shim | dropped, spec re-added natively | n/a | dropped, spec re-added natively |
| Speak `version` `"v1"` → open string | **identical change**, shipped as "No breaking changes" | getter already `Optional<String>` | same change, shipped as non-breaking |
| Generator type renames (`DeepgramSpeakProviderModel` → `DeepgramModel`, `Anthropic*`, `Google*`, `Groq*`, `Cartesia*`) | aliased back | renamed back | **N/A** — JS models these as nested `Deepgram.Model` via `enableInlineTypes`, so it never had those top-level names (0 occurrences on `main` and `HEAD`) |

The speak-provider `version` widening is part of the deliberate **Flux non-breaking reshape** (deepgram-docs#1064) and is byte-for-byte the same change Python ships as non-breaking (`Optional[Literal["v1"]]` → `Optional[str]`). It cannot be shimmed while keeping Flux TTS v2 usable — any type admitting `"v2"` is by definition not assignable to `"v1" | undefined`, and TypeScript has no divergent read/write property types. Treated as non-breaking here for consistency with Python and with this SDK's open-enum convention (89 files already use `| string`; the generator runs `enableForwardCompatibleEnums: true`).

**Java runtime defect checked here and not applicable:** java #89 found version-less providers being silently dropped (`{"provider":null}`) via Jackson's `defaultImpl = _UnknownValue`. JS performs no runtime union resolution — `sendJson` is `toJson(payload)` straight to the socket — and the regression test asserts a version-less provider serializes to exactly `{"type":"deepgram","model":"flux-general-en"}` with no injected `version` key.

**Spec provenance:** per java #89 this regen was cut from the docs `sdk-gen-2026-07-31` branch (deepgram-docs #1069, #1064, #1066, #1063, #1036, #1047), not docs `main` — consistent with `originGitCommitIsDirty: true` in `.fern/metadata.json`. If any of those docs PRs change before merging, this regen needs to be re-cut.

## Examples: `ForceEndTurn` is ahead of production (CI green, but blind to it)

`examples/38-transcription-flux-eot.ts` now demonstrates the new `ForceEndTurn` message and the `EndOfTurn` `trigger` field. The `examples` job reports **38/38 passed**, but that signal is weaker than it looks: `make examples` decides pass/fail purely on process exit code, and this example catches errors and calls `process.exit(0)`. Running it live today, production `/v2/listen` still rejects the new message:

```
Error UNPARSABLE_CLIENT_MESSAGE: Could not deserialize last text message:
unknown variant `ForceEndTurn`, expected `CloseStream` or `Configure`
```

So the SDK is correctly ahead of the server. This resolves itself when the API rollout and spec land (this PR is gated on those anyway) — no action needed, but don't read "38/38" as proof the new message works end to end.

**Follow-up worth filing:** `make examples` is exit-code-only, so an example can fail semantically while reporting PASS. Asserting on output (as the browser suite does via `hasSuccessOutput`) would make this catchable.

The same commit also fixes three pre-existing defects in that example, all found by running it live:
- `case "EagerEndOfTurn"` / `case "TurnResumed"` were **dead branches** — those are `event` values inside `TurnInfo`, not top-level message types (the v2 union is only `Connected` / `TurnInfo` / `ConfigureSuccess` / `ConfigureFailure` / `Error`). Now branches on `data.event` and surfaces the whole turn lifecycle.
- `case "FatalError"` **never matched** — `ListenV2FatalError` has `type: "Error"` — so fatal errors fell through to `default` and the connection was never closed.
- A pending `CloseStream` timeout could fire after the socket closed, throwing `Socket is not open.` and crashing the process. Now guarded on a close flag.

## Validation
- `make build` ✅
- `make test` ✅ (unit 586 + wire 140)
- `make typecheck-tests` ✅
- `make lint` ✅ (393 non-blocking warnings)


dg-coreylweathers (dismissed):
Approving the pre-regen prep. Verified this is the unfreeze step, not the regen output — 12 temporarily-frozen files copied to .bak siblings with .fernignore repointed to the .bak paths.

Checks (all pass):
- All 12 .bak files are byte-exact copies of their patched originals on main (diffed each pair).
- All temporarily-frozen entries swapped, none missed; permanently-frozen entries (CustomClient.ts, index.ts, transport.ts, scripts/tests/config) correctly untouched.
- Originals preserved (not deleted) so the generator can overwrite them; additions-only diff.
- stt_latency shim (AgentV1LatencyReport.ts) correctly included — the highest-stakes patch to protect this cycle.
- No build-gate impact: .bak is outside every tool glob (tsc/vitest/biome). The full make build/test/lint gate belongs on the follow-up regen-output PR.

One optional nit (separate commit, outside this diff): AGENTS.md's 'temporarily frozen files' list is stale — it enumerates 11 but this correctly swaps 12; AgentV1LatencyReport.ts (added to .fernignore in the 2026-07-20 stt_latency cycle) was never added to the AGENTS.md roster. The .fernignore comment is correct so nothing was missed here, but worth syncing so a future regen doesn't skip that patch.

Leaving the merge to you. Real review lands on the regen-output PR.

dg-coreylweathers (approved):
Content approved — Fern regen verified locally (typecheck-tests ✅, unit 588 ✅, wire 140 ✅; regression guard confirmed load-bearing; all 11 frozen patches byte-identical to main; additive surface with no removed exports/enums; the one breaking change correctly shimmed as a merged shape). S1 resolved across PR body, .fernignore, and AGENTS.md.

⚠️ Hold merge until the gate lifts: prod /v2/listen still rejects ForceEndTurn and the regen was cut from the unmerged docs branch sdk-gen-2026-07-31. Merge once the API rollout + spec land; re-cut if any of those docs PRs change first.

Optional nit (N1): examples/38-transcription-flux-eot.ts:85 logs "Fatal Error:" on the general case "Error:" branch.

PR #529, opened 2026-08-06
fix(websocket): default binaryType to arraybuffer for Bun

### LINKED ISSUE #525: Streaming WebSocket clients fail under Bun: SDK sets binaryType="blob" which Bun's ws rejects
## Summary
Every WebSocket-based client (`listen.v1`/`listen.v2`/`speak.v1`/`speak.v2`/`agent.v1`) throws immediately when run under **Bun**, because the SDK's WebSocket layer sets `binaryType = "blob"`, which Bun's bundled `ws` implementation rejects. The same code paths work correctly under Node.js.

## Environment
- `@deepgram/sdk` **5.7.0**
- Bun **1.3.14**
- Deepgram-facing (outbound) WebSocket — i.e. the SDK connecting to `api.deepgram.com` / `agent.deepgram.com`.

## Error / stack
```
error: Invalid binaryType: blob
  at binaryType (ws:258:18)
  at .../@deepgram/sdk/dist/esm/core/websocket/ws.mjs:373:18
```
This surfaces the moment the Deepgram socket opens, so no messages are ever exchanged; in a server it takes down the process via an unhandled rejection.

## Root cause
`core/websocket/ws.mjs` sets `socket.binaryType = "blob"`. Node's `ws` tolerates this, but Bun's `ws` shim only accepts `"nodebuffer" | "arraybuffer" | "fragments"` and throws on `"blob"`.

## Repro
Run any streaming example under Bun, e.g.:
```ts
import { DeepgramClient } from "@deepgram/sdk";
const dg = new DeepgramClient({ apiKey: process.env.DEEPGRAM_API_KEY });
const conn = await dg.listen.v1.createConnection({ model: "nova-3", encoding: "linear16", sample_rate: 16000 });
conn.on("open", () => console.log("open"));
conn.connect();
await conn.waitForOpen(); // throws: Invalid binaryType: blob
```
`bun run repro.ts` throws; the identical script under `node` works.

## Impact
Bun is a supported/target runtime for several Deepgram starter apps; all streaming starters are currently unusable on Bun.

## Suggested fix
Detect the runtime (or feature-detect accepted values) and set `binaryType = "arraybuffer"` (or `"nodebuffer"`) when `"blob"` is unsupported, rather than unconditionally `"blob"`.

edenbuilds:
## Summary
- Default WebSocket `binaryType` to `"arraybuffer"` instead of `"blob"` so Bun's `ws` (which rejects `"blob"`) can open streaming connections.
- Same default on the transport adapter in `CustomClient` for consistency.

## Test plan
- [x] Simulated Bun setter: `"blob"` → `Invalid binaryType: blob`; `"arraybuffer"` → accepted.

Fixes #525

### DIFF
--- a/src/CustomClient.ts
+++ b/src/CustomClient.ts
@@ -461,7 +461,7 @@ class TransportWebSocketAdapter {
     private _retryCount = -1;
     private _shouldReconnect = true;
     private _connectLock = false;
-    private _binaryType: BinaryType = "blob";
+    private _binaryType: BinaryType = "arraybuffer";
     private _closeCalled = false;
     private _messageQueue: DeepgramTransportMessage[] = [];
     private _connectTimeout: ReturnType<typeof setTimeout> | undefined;
--- a/src/core/websocket/ws.ts
+++ b/src/core/websocket/ws.ts
@@ -83,7 +83,9 @@ export class ReconnectingWebSocket {
     private _connectTimeout: any;
     private _shouldReconnect = true;
     private _connectLock = false;
-    private _binaryType: BinaryType = "blob";
+    // Prefer ArrayBuffer: Bun's `ws` rejects `binaryType = "blob"`, while
+    // ArrayBuffer is accepted by browsers, Node `ws`, and Bun.
+    private _binaryType: BinaryType = "arraybuffer";
     private _closeCalled = false;
     private _messageQueue: ReconnectingWebSocket.Message[] = [];
 


GregHolmes:
Thanks for picking this up, @edenbuilds — you correctly identified `binaryType = "blob"` as the culprit. I've opened #535 to carry this forward with two adjustments: the default is set in the frozen CustomClient wrapper so it survives Fern regeneration, and inbound binary is normalized back to a Blob so the `on("message")` payload type stays unchanged (keeping it non-breaking). You're credited as co-author on the commit. Closing this in favor of #535.

PR #531, opened 2026-08-11
revert: listen v2 ForceEndTurn + redact, diarization info, Flux TTS voices (#526)

GregHolmes:
Reverts the listen v2 regen merge (`5b8de76`, PR #526: ForceEndTurn + redact, diarization info, Flux TTS voices) off `main`.

Mirrors the equivalent revert in the Python SDK (deepgram-python-sdk#757).

## Why
This work isn't ready to ship yet, and other priorities need to release from `main` without it. Since it's already on `main`, any release cut from `main` would publish it.

## How
`5b8de76` is a merge commit (2 parents), so this is a single `git revert -m 1` that undoes everything PR #526 brought in — the 8 branch commits `76fcf17..b891f41` (unfreeze, regeneration, patch re-application, the `AgentV1UpdateListen` provider back-compat fixes, and the docs/example updates).

Verified exact: the resulting tree is **byte-identical** to `d330c39` (`chore(main): release 5.7.0`), the state of `main` immediately before the merge.

## Verification
- `make lint` (includes `typecheck-tests`) — clean
- `make build` — CJS + ESM built, ESM validation passed
- `make test` — **38 unit files / 580 tests** and **29 wire files / 139 tests**, all passing
- Working tree clean after the run

## Restoring later
The reverted work is preserved in the merged PR #526 and its branch commits; bring it back with a revert-of-this-revert, or re-run the generator when the feature set is ready to ship.

## Follow-up: release notes
This revert does **not** remove the `feat(regen)` commit for #526 from git history, and release-please rebuilds its notes from every commit since `v5.7.0`. So the regenerated release PR for 5.8.0 will still list the reverted features (`sendForceEndTurn()`, `redact`, `TurnInfo.trigger`, `diarize_info`, the 12 Flux voices) under **Features**, alongside a new **Reverts** entry — advertising APIs that aren't in the tarball. (Confirmed on the Python side: #757 merged, yet its regenerated release PR still listed the reverted feature.)

Before merging the release PR, hand-edit `CHANGELOG.md` on `release-please--branches--main--components--sdk` to drop the #526 feature bullets (keep the `Reverts` entry), mirror the edit into the PR body, then merge. release-please overwrites that branch on every push to `main`, so make this the **last action** before merge. Apply the identical fix on the Python side (deepgram-python-sdk).

PR #532, opened 2026-08-11
feat(regen): speak v2 barge-in and speed/expressivity controls, listen v2 redaction

GregHolmes:
JS SDK regeneration from the `jherlihy/flux-tts-ga` docs branch. Branched off `gh/revert-listen-v2-regen-2026-08-11` (#531) so it lands on a clean base rather than on the unreleased listen-v2 work.

Mirrors the Python regen (deepgram-python-sdk#758) — same spec baseline, same feature set.

## What this adds
**Speak v2:** `SpeakV2Interrupt` (barge-in, optional `playback_offset`), `SpeakV2Configure` with `ConfigureSuccess`/`ConfigureFailure`, `SpeakV2SpeechInterrupted`, plus `SpeakV2Expressivity` and `SpeakV2Speed`. New socket methods `sendInterrupt()` / `sendConfigure()`.

**Listen v2:** `ListenV2Redact` (`numbers`, `aggressive_numbers`).

**Not included:** `ForceEndTurn` and diarize info — verified absent (0 refs). Removed by the revert; this spec baseline predates them, matching Python.

## Manual patch reconciliation
**Re-applied (11):** `version.ts` release-please marker; `BaseClient` version single-sourcing; `Client.ts` `fetch()` default host; `package.json` devDeps Fern strips; `biome.json` `useIgnoreFile`; `CreateKeyV1RequestOne` alias; `AgentV1Settings` interface/namespace form (kept the generator's new `thought_signature`); `LanguageHint` namespace + deprecated `language_hint`; `ListenV2CloseStream.Type`; optional `create(request?)`.

**Dropped (1):** `AgentV1LatencyReport` — `stt_latency` is emitted natively again, so it's unfrozen and removed from `.fernignore` + AGENTS.md.

**Re-created (1):** the `AgentV1UpdateListen` provider shim the revert removed. The regenerated union breaks existing TS callers three ways — `version` becomes required (TS2322), `provider.model` widens to `string | undefined`, and reads of V2-only fields (`eot_threshold`, `language_hints`, …) fail with TS2339 because union access requires the field on every arm. No union can fix the third, so the field stays a merged shape derived from `DeepgramListenProviderV2`. Compile-compat only; the wire payload is unchanged. Now frozen with AGENTS.md documentation.

Also synced `pnpm-lock.yaml`, and added the previously-undocumented `AgentV1LatencyReport` entry to the AGENTS.md ledger during prep (pre-existing drift). `.fernignore` ↔ AGENTS.md now has zero drift both ways.

## New test coverage (13 tests, mutation-verified)
The regen shipped public surface with **zero** coverage — no test file referenced any of the 9 new symbols.

- **`tests/unit/regen-constraints.test.ts` (+11)** — placed here deliberately because it's inside `tsconfig.typecheck.json`, so the type assertions actually gate. Covers Interrupt (bare + `playback_offset`), Configure, ConfigureSuccess, ConfigureFailure (+ open-ended `Code`), SpeechInterrupted with the new required `breaks_applied`, its twin `SpeakV2SpeechMetadata.breaks_applied` (previously gated only from the uncompiled wire tree, so dropping it passed the gate — now pinned here too), `ListenV2Redact`, and the `AgentV1UpdateListen` provider back-compat (all three break modes — this shim was silently lost once, so it is now pinned).
- **`tests/wire/websocket/speak-v2-tts.test.ts` (+2)** — real mock-server round trips asserting the exact JSON on the wire for `sendConfigure`→`ConfigureSuccess` and `sendInterrupt`→`SpeechInterrupted`, plus `ConfigureFailure` delivery.

Mutation-checked: removing `breaks_applied` from either `ControlsApplied` twin → 2 typecheck errors each (the metadata twin was **0** before this gate was added); removing the provider shim → 5.

## Bug found and fixed
The speak-v2 wire test constructed a `SpeakV2SpeechMetadata` **without** the newly-required `breaks_applied` — a real type error this regen introduced. It went unnoticed because `tsconfig.typecheck.json` intentionally excludes the wire tree and vitest strips types, so that file's "typed payloads surface schema drift as a compile error" guarantee did not hold for it. Field added.

## Compatibility: no breaking changes vs v5.7.0
The revert-branch tree is byte-identical to the `v5.7.0` tag, so "vs last release" and "vs the revert" are the same comparison.

- **0** removed public exports (480 → 493, purely additive) and **0** deleted `src/` files
- no `optional → required` field changes; **two existing fields widen on the read path** — `Deepgram.version` (`"v1"` → `string`) and `Google.version` (`"v1beta"` → open `Google.Version`). Both are safe for callers *setting* the field; a consumer assigning them into a narrowed literal binding (e.g. `const v: "v1" | undefined = provider.version`) must widen the annotation.
- the provider back-compat is proven by explicit `--strict` assertions now pinned in the test suite

Only nit: `SpeakV2SpeechMetadata.ControlsApplied` gains a required `breaks_applied`. It's a server-emitted message you read rather than construct, so it's safe on the read path.

## Verification
- `make typecheck-tests` — 0 errors (**was failing before the patches were re-applied**, on `CreateKeyV1RequestOne` and the `Agent.Context/Listen/Think/Speak` namespaces)
- `make lint` — exit 0; **396** biome warnings, up from 393 on `v5.7.0`. The 3 new ones are `any` usages in the added WebSocket wire test; no errors.
- `make build` — CJS + ESM, ESM validation passed
- `make test` — **590 unit / 141 wire**, all passing
- `make test-esm` — 5 passing
- **25 of 26 non-management examples verified green** against the live API **as of `970c907`**. `26-transcription-live-websocket-v2` is a deliberately persistent keepalive demo with no exit condition, so it hits a timeout by design. The 12 management/on-prem-credential examples were intentionally skipped. **Example 37 (`37-text-to-speech-streaming-flux.ts`) was materially extended afterward** (barge-in path, `sendConfigure`, new message handlers) and its added field usage was verified by reading against the generated types; it has **not** been re-run live — a tracked follow-up (needs a paid endpoint).

## Known issue carried from the spec (applies to Python too)
`SpeakV2Speed` and `SpeakV2Expressivity` lose their spec enum constraints in codegen (numeric enums generate as bare `number`; string enums keep their `Literal` unions), so out-of-range values are only caught by the server. Pinned in the new tests so a future generator that narrows them is noticed.


dg-coreylweathers (dismissed):
**Content approved — but hold the merge until the `expressivity` question is settled.** Flagging both halves separately because they are not the same call.

**First, a correction to something I raised earlier:** I initially flagged this as regenerated from a stale spec commit. That was wrong — I misread the GitHub commits API as newest-first when it returns oldest-first. `03f0677` **is** the head of deepgram-docs#1090 (confirmed via `.head.sha`, and `compare/ba0643513c...03f06776bb` reports `ahead_by: 3, behind_by: 0`), and the `fix` and GA-examples commits are ancestors already baked into this tree. Nothing to re-regenerate on that account. Apologies for the noise.

**The regen mechanics and patch reconciliation check out.** I verified rather than took on trust:

- The `AgentV1UpdateListen` shim is correctly re-created — swapping the merged shape back for the generated union produces exactly the 5 errors you documented, mapping 1:1 to the three break modes.
- `speed` and `expressivity` genuinely reach the upgrade URL through the hand-written wrapper. This was the highest-risk check, since `CustomClient` hand-rolls that URL — `buildQueryParams` is a denylist over transport-only keys, not an allowlist, so new params pass through. Runtime-confirmed.
- `WrappedSpeakV2Socket extends SpeakV2Socket`, so `sendInterrupt`/`sendConfigure` inherit correctly.
- The revert branch is byte-identical to the `v5.7.0` tag, which makes "vs the revert" and "vs last release" the same comparison — so removing the unreleased #526 surface is correctly non-breaking.
- Locally: `make typecheck-tests` 0 errors, `make test` all passing, `make lint` exit 0.
- Python's websocket credential leak does **not** replicate here — `DeepgramError` never receives request headers, and the debug path never logs headers or the URL.

**Blocking on merge: `expressivity`.** It ships typed and documented on both the streaming connect path and the batch REST path, and the server rejects every value including the documented default `0` (I confirmed `expressivity=0` does serialize onto the query string, so nothing softens it). Beyond your own staging finding, Rohit posted in `#project-flux-tts` about 25 minutes after your last update here: *"we are at high risk for this not shipping w GA because of work we are doing to triage/fix the hallucinations in staging right now... reliability is P0."* So the open question is no longer "when does it deploy" but "does it ship at all."

Cleanest fix, and it fixes JS, Python, and Java at once: get `expressivity` pulled from spec #1090 before it merges. Fern generates only what the spec declares, so no `.fernignore` patch and no hand-edited generated files. It is also nearly free — `expressivity` is the single newest commit on that branch (`03f0677`, +8 lines of overrides plus one OpenAPI line), cleanly separable from the GA spec work that landed the day before. Needs Jane and Rohit.

**One should-fix worth doing regardless** — the field you call out as the only compatibility concern is the one thing with no gate:

```
tests/unit/regen-constraints.test.ts
Add a SpeakV2SpeechMetadata literal next to the existing SpeechInterrupted one.
```

Deleting `SpeakV2SpeechMetadata.ControlsApplied.breaks_applied` currently produces **0** typecheck errors — its only reference is in `tests/wire`, which `tsconfig.typecheck.json` excludes. The `SpeechInterrupted` twin produces 2. That is the same blind spot that hid the bug you fixed in this PR, so it can recur next regen.

**Smaller items:** lint is 396 on this branch vs 393 on `v5.7.0` (all 3 new ones in the added wire test, so not "identical count"); `reference.md` has 0 mentions of `speed`, `expressivity`, `Interrupt`, or `Configure` (Python fixed this); no barge-in example yet (Python added one, and the offset semantics are easy to get wrong from types alone); and the `requests.list` wire test lost all 10 query params — the same upstream regression you restored on the Python side. Also worth a sanity check: the compatibility note says "no read-widening on existing fields," but `Deepgram.version` and `Google.version` both widened from a literal to open `string`.

Full detail in the review file.

dg-coreylweathers (approved):
Approved. The regen mechanics, patch reconciliation, and the B1 copy bridge all check out — I verified rather than took on trust.

**Verified independently:**
- `d44f16f`'s bridge is byte-faithful to spec #1090 across all five files, including `SpeakV2Request.ts` on the batch path.
- S1's gate is now symmetric: mutation-tested both twins, 2 typecheck errors each (the metadata twin was 0 before).
- Locally on `e70ef38`: `make typecheck-tests` 0 errors, 591 unit / 141 wire passing, `make lint` exit 0 at 396.
- Example 37 **is** verified live — CI's `examples` job runs `make examples` with a real key against production, no timeout wrapper, `exit 1` on any failure, and it passes on this head. The PR body's "not re-run live" caveat is over-conservative; that follow-up can be dropped.

**Deferred by me, not dismissed — must land before the release-please PR merges:**
- **B2** — `SpeakV2Expressivity` ships with no Beta/hallucination-risk warning. The GA eval measured `+2` at ~3.4x neutral warble and `-2` degrading speed/pause-control accuracy; the spec carries that language and the SDK docstring is where developers pick the value. Fix separately or let the post-#1090 regen reproduce it.
- **B3** — the post-#1090 regen is a planned breaking change, not a cleanup: streaming `SpeakV2Speed` becomes a string enum, so `speed: 1.05` in example 37 and `regen-constraints.test.ts` must be updated in the same commit, and the type change needs to reach the release notes. An AGENTS.md note recording which of the 8 unfrozen hand-edits get reproduced vs reverted would be cheap insurance.

**One thing to fix before you hit merge:** the PR body still says `expressivity` is not deployed and 400s on every value. This squash-merges, so that lands in the commit message and release notes. Replacement copy is in my review comment.

Fast-follows, agreed as tracked: S2 (`requests.list` query assertions), S4 (`reference.md` — note a regen won't fix this, TS generator is REST-only), N2 (Fern defect list).

Also worth flagging: JS is now ahead of Python on launch-day accuracy — #758's last commit predates both spec fixes and has no equivalent bridge, so Python still ships the "controls work" copy.

PR #535, opened 2026-08-13
fix(websocket): default binaryType to arraybuffer for Bun compatibility

### LINKED ISSUE #525: Streaming WebSocket clients fail under Bun: SDK sets binaryType="blob" which Bun's ws rejects

GregHolmes:
## Summary

Every WebSocket streaming client (`listen`/`speak`/`agent`) throws immediately under **Bun**, because the SDK's WebSocket layer sets `binaryType = "blob"`, which Bun's bundled `ws` rejects (`Invalid binaryType: blob`) the moment the socket opens. Node tolerates `"blob"`; Bun does not.

This defaults the socket's `binaryType` to `"arraybuffer"` — accepted by browsers, Node `ws`, and Bun alike.

## Why this shape

- **Regen-durable, no freeze needed.** The default is set in `CustomClient.ts` (the sole `ReconnectingWebSocket` factory, already frozen in `.fernignore`), *not* in the generated `core/websocket/ws.ts`. So Fern regeneration can't silently revert it, and no generated runtime file has to be frozen.
- **Non-breaking.** `"arraybuffer"` changes the raw inbound binary type from `Blob` to `ArrayBuffer`. To preserve the existing public contract, `setupBinaryHandling` now wraps inbound binary back into a `Blob` before delivery — so `on("message")` hands consumers a `Blob` on every runtime, exactly as before. No consumer-visible type change, no major bump.

## Tests

- **Unit:** asserts the `"arraybuffer"` default across `speak`/`listen`/`agent` v1 via the real client factory; corrects a stale copy-test.
- **Wire:** tightened to pin inbound audio as a `Blob` (was tolerant of `ArrayBuffer || Blob`).
- Full unit (594) + wire (141) suites green.
- **Live e2e (Node + Bun):** speak.v1 TTS delivers real audio as `Blob`, and listen/speak/agent all open without the crash on both runtimes. Verified via negative control that reverting the fix reproduces `Invalid binaryType: blob` under Bun.

Fixes #525

Co-authored-by: edenbuilds <279970382+edenbuilds@users.noreply.github.com>

### DIFF
--- a/examples/09-voice-agent.ts
+++ b/examples/09-voice-agent.ts
@@ -41,9 +41,9 @@ async function voiceAgent() {
             // Check message type
             if (data.type === "ConversationText") {
                 console.log("Conversation text:", data);
-            } else if (typeof data === "string") {
-                // Audio data comes as string
-                console.log("Audio received (length):", data.length);
+            } else if (data instanceof Blob) {
+                // Audio data is delivered as a Blob
+                console.log("Audio received (bytes):", data.size);
             } else {
                 console.log("Message:", data);
             }
--- a/examples/36-agent-inject-message.ts
+++ b/examples/36-agent-inject-message.ts
@@ -67,7 +67,8 @@ async function agentInjectMessage() {
                 console.log("Agent started speaking");
             } else if (data.type === "AgentAudioDone") {
                 console.log("Agent finished speaking");
-            } else if (data instanceof Uint8Array || Buffer.isBuffer(data)) {
+            } else if (data instanceof Blob) {
+                // Audio data is delivered as a Blob
                 console.log("Audio data received");
             } else {
                 console.log("Message:", data);
--- a/src/CustomClient.ts
+++ b/src/CustomClient.ts
@@ -980,9 +980,13 @@ function setupBinaryHandling(
                 // If JSON parsing fails, pass the raw string
                 eventHandlers.message?.(event.data);
             }
-        } else {
-            // Binary data - pass through as-is
+        } else if (event.data instanceof Blob) {
+            // Already a Blob - pass through as-is.
             eventHandlers.message?.(event.data);
+        } else {
+            // Binary arrives as an ArrayBuffer (the socket uses binaryType "arraybuffer").
+            // Wrap it in a Blob so consumers always receive a Blob, regardless of runtime.
+            eventHandlers.message?.(new Blob([event.data as BlobPart]));
         }
     };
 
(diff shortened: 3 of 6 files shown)


dg-coreylweathers (approved):
Reviewed — fix mechanism verified end-to-end (root cause, regen-durable placement in the frozen CustomClient.ts wrapper, override timing, all 5 streaming clients covered, public Blob contract preserved). Ran the actual test/lint gates myself: unit 596 passed, wire 141 passed, tsc clean, biome lint/format clean. One caveat noted in the review: no Bun runtime available to independently reproduce the crash/fix, so that part is author-attested.

lukeocodes (approved):

PR #539, opened 2026-08-31
feat(examples): add production-grade live reconnection example

dg-coreylweathers:
## What

Adds `examples/41-transcription-live-reconnect.ts` - the canonical production-grade reconnection example for `/v1/listen` streaming - plus Makefile (`make example-41`, help text, runner) and `examples/README.md` wiring.

## Why

Real-world streaming connections drop, and we do not have a runnable reference showing the full recovery pattern. This example is written to back the production streaming guide on developers.deepgram.com (`/docs/speech-to-text/streaming/guides/production` - page not published yet).

## What it demonstrates

- Full-jitter exponential backoff with a retry cap (5 attempts, 1s base, 10s cap)
- Reconnect-worthy close codes (1006, 1011, 1012/1013, 4xxx) distinguished from normal/fatal closure (1000, 1002/1003/1007/1008)
- Audio buffered while disconnected (30s cap, oldest dropped) and flushed after reconnect
- Original connection options re-applied on every reconnect
- Timestamp realignment across connections (each new connection restarts at 0s)
- KeepAlive every 5s and a clean CloseStream -> Metadata -> code-1000 shutdown
- Raw linear16 with explicit encoding params, so mid-stream resumed audio stays decodable (container headers only exist at stream start)

Uses only the SDK client for Listen v1; transport-level retries are disabled (`reconnectAttempts: 0`) so the app owns the reconnect policy. By default the example forces one mid-stream disconnect (close 4000) to demonstrate recovery; `SIMULATE_DROP=0` streams without it.

## Validation

Deterministic tests cover buffer overflow timestamp continuity, the forced reconnect path, and clean or failed shutdown. `biome check` is clean.

## Test

```bash
DEEPGRAM_API_KEY=<key> make example-41
```


GregHolmes (changes requested):
# Review: deepgram-js-sdk #539, feat(examples): add production-grade live reconnection example
Classification: mixed (code-led; docs lens on examples README)
Verdict: request-changes

## Intent
Add a runnable, canonical `/v1/listen` reconnection example that owns retry policy, preserves recoverable audio, reuses transcription options, and presents a continuous transcript timeline. The PR description is the only supplied intent context; no linked issue or Slack thread was available.

## Blocking
- [B1] `examples/39-transcription-live-reconnect.ts:157` Dropped audio is omitted from the transcript timeline.

  **Summary:** The example intentionally discards the oldest buffered audio after a disconnect exceeds 30 seconds, but calculates the next connection's timestamp offset from `deliveredBytes` only. If eight seconds were delivered before a disconnect and ten buffered seconds were discarded before reconnection, the next retained audio starts at source time 18 seconds but is logged as starting at eight seconds. This makes the advertised continuous timeline incorrect precisely in the long-outage case the buffer cap handles.

  **Expected:** Every transcript timestamp after a reconnect must represent the source-audio position of the bytes actually sent to that new connection, including time skipped by intentionally discarded buffered audio.

  **Observed:** `timestampOffsetSec = deliveredBytes / BYTES_PER_SECOND` at `examples/39-transcription-live-reconnect.ts:300` excludes every byte removed in `bufferChunk()`, so post-overflow transcript timestamps overlap or lag the already logged timeline.

  **Recommended fix:** Track discarded bytes and include them in the reconnect offset. For example, add `let droppedBytes = 0` beside `deliveredBytes`, add `droppedBytes += dropped.length` immediately before or after `bufferedBytes -= dropped.length` at line 159, and use `(deliveredBytes + droppedBytes) / BYTES_PER_SECOND` at line 300. Add a deterministic test that exceeds the buffer cap and asserts the first post-reconnect timestamp starts after both sent and discarded audio.

## Should-fix
- None.

## Nits
- [N1] `examples/README.md:19` The new catalog entry names a nonexistent JavaScript file.

  **Summary:** The newly added example is TypeScript, but the README points developers to `39-transcription-live-reconnect.js`. A developer following the catalog entry cannot open or run that path directly.

  **Expected:** The examples catalog must name the checked-in file a developer can execute through the documented tooling.

  **Observed:** `examples/39-transcription-live-reconnect.ts` exists, while `examples/README.md:19` names the `.js` variant.

  **Recommended fix:** Change the entry to `**39-transcription-live-reconnect.ts**`. The older catalog entries have the same stale extension pattern, but this new entry should not extend it.

## Verified (evidence)
- The SDK uses the required two-stage WebSocket flow: create the connection, register handlers, call `connect()`, then await `waitForOpen()` (`examples/39-transcription-live-reconnect.ts:222-273`; `src/api/resources/listen/resources/v1/client/Socket.ts:91-129`).
- `reconnectAttempts: 0` disables only the SDK transport retry while each app-owned retry creates a new socket with the original connection options (`examples/39-transcription-live-reconnect.ts:222-229,277-305`; `src/core/websocket/ws.ts:357-408`).
- Normal/fatal close handling, KeepAlive, and CloseStream-to-Metadata shutdown paths are explicitly implemented (`examples/39-transcription-live-reconnect.ts:184-189,247-269,316-334`).
- `git diff --check origin/main...HEAD` passed.
- `pnpm exec biome check examples/39-transcription-live-reconnect.ts Makefile examples/README.md` passed.
- `pnpm exec tsc --noEmit --allowJs --checkJs false --target ES2022 --module commonjs --moduleResolution node --types node examples/39-transcription-live-reconnect.ts` passed.
- `pnpm exec vitest --project unit --run` passed: 38 files and 591 tests.
- `make lint` passed locally, including the type-check gate; it reported 396 pre-existing repository warnings. GitHub reports all PR checks passing, including the examples job.

## Needs human
- The claimed three successful production runs are not independently reproducible without an API key, and the default forced-drop run cannot reach the 30-second buffer-overflow path because `spacewalk.wav` is shorter than the buffer cap. The requested deterministic overflow test is needed before treating the timestamp-continuity claim as verified.

## Developer-facing messaging
- The example is valuable because it shows that reconnecting live STT requires application-owned buffering and a new connection with the same options. Once B1 is fixed, describe timestamp continuity as applying across both ordinary reconnects and the deliberate audio-loss policy; otherwise developers will receive timelines that are wrong after a prolonged outage.

GregHolmes (approved):

PR #541, opened 2026-09-03
fix: restore wire imports after test failures

GregHolmes:
## Summary

- Run wire-import restoration from an exit trap so `make test` cleans up after success, test failure, interruption, or termination.
- Preserve the original test failure status.
- Surface a cleanup failure when the tests themselves passed.

## Verification

- `make build`
- `make test`
- `make lint`
- Forced `pnpm` to exit nonzero after the import rewrite and confirmed `make test` also failed while `git diff --exit-code -- tests/wire` remained clean.

No live API verification applies because this change only controls local test-runner cleanup.

Follow-up from the PR #538 agent-instructions review and PR #514 coverage review.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -320,10 +320,20 @@ build:
 	node scripts/validate-esm-build.mjs
 
 test:
-	node scripts/fix-wire-test-imports.js
-	pnpm exec vitest --project unit --run
+	@cleanup() { \
+		test_status=$$?; \
+		trap - 0 INT TERM; \
+		node scripts/revert-wire-test-imports.js; \
+		cleanup_status=$$?; \
+		if [ $$test_status -ne 0 ]; then exit $$test_status; fi; \
+		exit $$cleanup_status; \
+	}; \
+	trap cleanup 0; \
+	trap 'exit 130' INT; \
+	trap 'exit 143' TERM; \
+	node scripts/fix-wire-test-imports.js && \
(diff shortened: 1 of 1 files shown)


naomi-deepgram (approved):

dg-coreylweathers (approved):
Approve.

**What this PR does.** `make test` rewrites 22 wire-test import lines before running and swaps them back afterwards; on `main` the swap-back only ran when every test passed, so a failing or interrupted run left 22 modified files behind. This moves the restore into a shell exit trap so it runs on success, failure, Ctrl-C, and SIGTERM, and keeps the original test status. CI runs `make test` too.

**What I checked.**
- Does the restore run on every exit path? Yes. I ran an exact copy of the recipe under both macOS `sh` and `dash` (CI's shell) through pass, unit failure, wire failure, rewrite-step failure, restore-step failure, both failing at once, process-group Ctrl-C (child dies of SIGINT, child exits 1, child exits 0), and SIGTERM. Files were restored and the exit status was right every time; a test failure still wins over a restore failure.
- Does the real thing still work? Yes. `make test` with this Makefile: 591 unit + 141 wire tests pass, `tests/wire` clean afterwards, exit 0.
- Does it change which suites run? No. Unit failure still skips the wire suite.

Nit: the new `@` suppresses echo of the three commands before their output; drop it or add an `echo` per step if you want that visibility back.

PR #447, opened 2025-12-08
feat!: release new codegen architecture and interfaces

naomi-deepgram:
## Types of changes

What types of changes does your code introduce to the community JavaScript SDK?
_Put an `x` in the boxes that apply_

- [ ] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update or tests (if none of the other choices apply)

## Checklist

_Put an `x` in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before merging your code._

- [ ] I have read the [CONTRIBUTING](../CONTRIBUTING.md) doc
- [ ] I have lint'ed all of my code using repo standards
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation (if appropriate)

## Further comments

Closes #457 
Closes #454 
Closes #451 
Closes #435 
Closes #355 
Closes #366 
Closes #395 
Closes #426 
Closes #346 
Closes #294 
Closes #189 
Closes #254

PR #543, opened 2026-09-03
chore: remove stale frozen-file entries

GregHolmes:
## Summary

- Remove `.fernignore` entries for `webpack.config.js` and `MIGRATION_GUIDE_V4_TO_V5.md`, which no longer exist.
- Remove the corresponding stale entries from the permanent-freeze inventory in `AGENTS.md`.
- Leave all active permanent and temporary freezes unchanged.

## Verification

- Confirmed both paths are absent from the repository.
- Confirmed no references remain outside the entries removed here.
- `git diff --check` passes.

Follow-up from the PR #538 agent-instructions review.

### DIFF
--- a/.fernignore
+++ b/.fernignore
@@ -11,7 +11,6 @@ scripts/fix-wire-test-imports.js
 scripts/proxy-server.js
 scripts/revert-wire-test-imports.js
 scripts/validate-esm-build.mjs
-webpack.config.js
 
 # Tests
 tests/browser
@@ -47,7 +46,6 @@ tsconfig.typecheck.json
 
 # Other Stuff
 Makefile
-MIGRATION_GUIDE_V4_TO_V5.md
 docs
 .npmignore
 .gitignore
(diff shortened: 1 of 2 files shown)


naomi-deepgram (approved):

dg-coreylweathers (approved):
Approve.

**What this PR does.** Removes two `.fernignore` entries and their AGENTS.md bullets for files that no longer exist: `webpack.config.js` and `MIGRATION_GUIDE_V4_TO_V5.md`.

**What I checked.**
- Are both files really gone? Yes. `webpack.config.js` was deleted in #447 (2026-03-04), the same commit that added its ignore line, so the entry was stale from day one. The migration guide moved to `docs/Migrating-v4-to-v5.md` on 2025-12-28; `docs` stays frozen and the README already links the `docs/` paths.
- Anything else reference them? No. Repo-wide sweep hits only CHANGELOG history, a bundler comment in `CustomClient.ts`, and the AGENTS.md note about `webpack`'s peer ranges, which is still accurate because `webpack` and `ts-loader` remain Fern-emitted devDependencies.
- Only these four lines removed? Yes, neighbours intact.
- Merge with open #538, which also edits AGENTS.md? No conflicts in either order.

Watch item, not a blocker: if the Fern TypeScript generator emits a `webpack.config.js`, the next regen branch will now include one since the ignore no longer suppresses it. It would be visible in that PR's diff.

PR #553, opened 2026-09-14
test(browser): wait for transcription completion

GregHolmes:
## Summary

Wait for a non-empty prerecorded-file transcript rather than treating the immediate progress message as completion. Failure output now includes the page response for diagnosis.

## Validation

- `make build`
- Targeted browser test passed three consecutive runs
- `pnpm exec biome check tests/browser/05-transcription-prerecorded-file.test.ts`


dg-coreylweathers (approved):
**What this PR does.** Fixes the intermittent `test-browser` failures by waiting for each example page's real completion marker instead of its immediate progress line. The polling loop lives in `tests/browser/helpers.ts` (`waitForOutputMatching` / `waitForSuccessOutput`, 25 s deadline under the tests' 30 s timeout, page text included on timeout) and is applied to tests 04, 05, 08, 10, and 12. Test-only; nothing under `src/` changes.

**What I checked.**
- The three CI failures on `main` today all died at the old `expect(hasSuccess).toBe(true)` line in 05, 08, or 12; all three files are covered now.
- Pages 04, 10, 12 print nothing containing `✓` or "success" before the API answers, so the generic matcher cannot pass early.
- `hasSuccessOutput` is kept as a wrapper over the new `isSuccessOutput`, so the other browser tests are unchanged.
- `biome check` on the six files, repo-wide `biome lint` + `biome format`, and `tsc --noEmit` over `helpers.ts` and every `tests/browser/*.test.ts`: all pass.
- Live: three consecutive runs of 04+05+08+10+12 in the Playwright v1.59.1 Docker image against prod, 15/15 passed. CI `test-browser` and `examples` are green on this head.
- Title is `test(browser)`, so release-please emits no changelog bullet and no version bump.

Optional, no need for another round: 04 still treats "✓ Transcription: No transcript found" as success (05 excludes it); an API error now surfaces after the 25 s deadline rather than immediately; the PR body still describes only the 05 change.

PR #555, opened 2026-09-14
fix(socket): support multiple listeners per event and add off()

### LINKED ISSUE #554: Socket.on() silently overwrites previous handlers instead of adding a listener
Proposed changes

  on() on the streaming socket clients assigns a single callback per event instead of registering a list:

  ts
  public on<T>(event: T, callback: V1Socket.EventHandlers[T]): void {
      this.eventHandlers[event] = callback;   // overwrites, doesn't add
  }

  A second .on("message", ...) call silently discards the first — no warning, no off() to unsubscribe either. Proposed: back eventHandlers with an array per event, and
  add off().

  Context

  Confirmed by repro — two message handlers registered, one event dispatched:

  handlers that fired = ["handler-B"]   (expected both A and B)
  off()/removeListener()?  undefined

  A method named on() is expected to behave like addEventListener, not a setter. Breaks any composition (logging layer + app logic, React re-mounts). Also a parity gap —
  Python SDK's sockets use a real EventEmitterMixin supporting multiple listeners.

  Possible Implementation
 src/CustomClient.ts already wraps every streaming socket and is excluded from Fern regen, so the fix lives there: array-backed eventHandlers, on() pushes, off()
  splices, dispatch iterates.

  Other information

  JS SDK v5.10.1. Same bug also leaks listeners in waitForOpen() — 3 calls leave 4 open listeners attached, no timeout either.

sudoKrishna:
What : 

Fixes #554 

  on() on the streaming sockets kept a single callback per event -- a second on("message", ...) silently displaced the first, with no off() to remove one. This adds
  multi-listener support and off() for all four events (open/message/close/error) across all five streaming sockets (agent.v1, listen.v1, listen.v2, speak.v1, speak.v2).

  - Re-registering the same callback reference for the same event is a no-op, matching EventTarget.addEventListener.
  - off() removes exactly the one it targets; removing a listener that was never added, or already removed, is a no-op.
  - Each dispatch runs over a snapshot of the listener list, so a listener that calls off() on itself (or another listener) mid-dispatch doesn't skip whoever comes after
  it — matches Node's EventEmitter.
  - Composes with existing async iteration (for await): the iterator is just one more listener now, not a special case.

  Where

  src/CustomClient.ts only -- the hand-written, .fernignore'd wrapper layer that already owns all five streaming sockets, so this survives Fern regeneration.

  Testing

  - New wire tests: tests/wire/websocket/multiple-listeners.test.ts (6 tests, real mock WS server) -- 
Screenshot From 2026-09-15 01-03-17
multiple listeners per event, off() mid-dispatch,
  duplicate-reference no-op, removing-unregistered no-op, and iterator/callback coexistence.
  - Full suite: 1397/1397 passing (make test equivalent via vitest; the 1 unrelated failing suite needs make build first per AGENTS.md, not touched by this change).
  - tsc --noEmit clean, including the public AsyncIterable*Socket types now exposing off().
  - biome lint/format: same pre-existing warning count as main, no new errors.


sudoKrishna:
Hi @GregHolmes  @deepgram-kiley @dg-coreylweathers , just following up on this PR. Would appreciate a review when you get a chance. Thanks!

GregHolmes:
Thanks for the work on this. I’m going to close this as superseded by the Fern 3.94.0 regeneration in #557.
Fern now generates off() on all streaming sockets, so that part is covered. The regen initially included the same accumulated-listener behavior, but we are retaining the existing replacement semantics for on() through v5 to avoid a behavioral breaking change for callers that re-register handlers to replace stale callbacks.
That means the multi-listener implementation here would conflict with the compatibility policy in the regeneration PR. We can revisit accumulated listeners in the next major release with explicit migration guidance.

PR #546, opened 2026-09-07
feat(regen): agent ForceEndTurn, agent speak expressivity, Flux TTS speed range 0.5-1.5

GregHolmes:
## Summary

- Regenerate the JavaScript SDK with Fern TypeScript generator 3.89.0.
- Add Agent `sendForceEndTurn()` and Deepgram speak-provider `expressivity`.
- Retain the released `SpeakV2Speed` string constants and `Aura2PerseoIt` constants as deprecated aliases for source compatibility.
- Reapply required compatibility and tooling patches.

## Production availability

Listen V2 ForceEndTurn requires server-side enablement. Deployments without the feature return `UNPARSABLE_CLIENT_MESSAGE` and close the connection.

## Unsupported voice compatibility

`aura-2-perseo-it` was never served by the API (it returns 400) and was removed from the spec. `Aura2PerseoIt` remains as a deprecated alias until the next major release so existing TypeScript callers keep compiling.

## Generator provenance

`originGitCommitIsDirty` is expected: Fern ran after an uncommitted `fern/generators.yml` change that switched only the JavaScript output mode and target branch to this regeneration branch. No API specification or JavaScript code-generation setting was uncommitted, so the generated SDK contract is unchanged by that marker.

## Validation

- `pnpm install --frozen-lockfile`
- `make build`
- `make lint`
- `pnpm exec vitest --project unit --run`
- non-management wire integration suite

The browser test 21 timeout is pre-existing: it also failed on `main` on 2026-09-04.

PR #557, opened 2026-09-16
feat(regen): agent function cancellation and defer-until-EOT

GregHolmes:
## Summary
- Regenerates the JavaScript SDK with fernapi/fern-typescript-sdk 3.94.0.
- Adds Agent FunctionCallCancelled events and the optional defer_until_eot function setting.
- Retains custom transport behavior, native Flux terminal-close handling, credential-safe passthrough requests, legacy Agent Settings and language-hint compatibility, Listen V2 compatibility, TTS model and speed aliases, and the single-source SDK version header.
- Preserves v5 replacement semantics for streaming socket on() handlers while retaining the new off() API and explicit shouldReconnect policy support.

## Validation
- make build: passed
- make test: 1,266 unit and 159 wire tests passed
- make lint: no errors; existing warnings only
- make typecheck-tests: passed
- pnpm tsx tests/manual/flux-terminal-close.ts: production Flux CloseStream closed with 1005 without reconnecting by default
- pnpm tsx tests/manual/agent-defer-until-eot.ts: production Voice Agent accepted a complete deferred-function schema

## Release note
- Fern generated source version 5.11.1. The feature-level PR will cause Release Please to cut the next minor version, 5.12.0; Release Please owns the published version stamp.


dg-coreylweathers (changes requested):
## What this PR does

Regenerates the SDK with Fern TypeScript 3.94.0, adding two Voice Agent features: the `FunctionCallCancelled` event and the `defer_until_eot` function setting. The generator release also changed socket `on()` from replacement to accumulation (five socket files frozen to keep v5 semantics while keeping the new `off()`), and added a `shouldReconnect` close policy, plumbed through every streaming client with a hand-patch in `src/core/websocket/ws.ts` so an explicit `reconnect()` cannot be vetoed by the policy.

The regenerated API surface is correct and fully covered. Both blocking findings are in the hand-written WebSocket layer.

## What I checked

- **Does a Flux stream stay closed when you close it?** No. Live against production with no `shouldReconnect` supplied: `sendCloseStream()` → close code 1005 → the SDK reconnected. Two `open` events for one intended session, and the process does not exit.
- **Does a throwing callback stay contained?** No. Reproduced on this branch: a throwing `message` callback prevents the async-iteration handler from running, and a throwing `close` callback leaves `for await` hanging with nothing logged.
- **Did the new Agent surface land correctly?** Yes. `FunctionCallCancelled` is in the agent response union, exported, and delivered end-to-end by a new wire test; `defer_until_eot` is typed and production Voice Agent returned `SettingsApplied` for it.
- **Did the manual patches survive?** Yes — spot-checked `SpeakV2Speed` constants, `Deepgram.Model.FluxReneeEn`, `DeepgramTimeoutError extends Error`, `CreateKeyV1RequestOne`, `Aura2PerseoIt`, the `SDK_VERSION` import in `BaseClient.ts`, and the Listen V2 `ForceEndTurn` caveat. `.fernignore` and `AGENTS.md` cover the newly frozen files; no leftover `.bak` files.
- **Everything else passed:** build, 1,423 unit and wire tests, and `tsconfig.typecheck.json` in a clean `node:24` container; `shouldReconnect` does not leak into the connection URL; `off()` correctly removes only the caller's handler; all nine CI checks green.

## Blocking

### [B1] The SDK opens a second connection after the caller asks to shut a Flux stream down

On `listen.v2`, `sendCloseStream()` is how a caller signals it is done. Deepgram closes the socket with no status, reported as code 1005, and the SDK's default policy is "reconnect on any code other than 1000" — so it reconnects.

**Expected:** after `sendCloseStream()`, the following close is final and the SDK does not reconnect, with no caller configuration — as it already does when a custom transport is in use, and as the Java SDK does in this same regen.

**Observed:** against production with no `shouldReconnect` supplied, two `open` events for one intended session, the second opened after the caller asked to stop. The caller is billed for a connection they explicitly closed and their process does not exit. Avoided only by writing a stateful `shouldReconnect` closure, which is what examples 26, 38, and 39 now do.

**Recommended fix.** Minimal: drop "native Flux terminal-close policy support" from the PR body and release notes, since the default path has no such policy, and document the `shouldReconnect` closure in the README. Robust (recommended): port the Java #110 policy to the default path, which is stricter than a boolean because it records *which socket* sent `CloseStream`:

```java
protected boolean shouldReconnectAfterClose(WebSocket webSocket, int code, String reason) {
    return super.shouldReconnectAfterClose(webSocket, code, reason)
            && (code != 1005 || closeStreamSocket.get() != webSocket);
}
```

That handles a rejected `CloseStream` and a fresh connection correctly, both of which Java covers with tests. Once the default path does this, the `closeStreamSent` closures can come back out of the three examples.

### [B2] One callback that throws stops the others, and `for await` then never finishes

Dispatch moved from a single stored handler to a loop over an array, and this PR removed the `try`/`finally` that isolated them. `installAsyncIteration` powers the `for await` support the README documents, and it works by placing its handler in that same array after the caller's.

**Expected:** a caller callback that throws affects only that callback; other registered handlers still run and `for await` still terminates when the connection closes. The replaced code did this deliberately — the deleted comment said both consumers must see every message regardless of which registered first.

**Observed:** a throwing `message` callback means `for await` never receives that message; a throwing `close` callback means the iteration handler never runs, so the loop is never told the connection ended and hangs forever. Both reproduced on this branch. Nothing is logged, so it presents as a process that stops producing output and will not exit.

**Recommended fix.** Minimal, one line in `installAsyncIteration`'s `on()` rebinding, so a throwing caller callback cannot starve iteration:

```ts
userHandlers.splice(0, userHandlers.length, handler as never, callback as never);
```

Robust: restore per-handler isolation in the dispatch loop of the five frozen socket files, wrapping each `handler(...)` call in `try`/`catch` and rethrowing asynchronously. Worth reporting upstream too — the unguarded loop is generator output.

## Should-fix

- **[S1]** `isTerminalClientMessage` recognizes only `{"type":"CloseStream"}`, but `speak.v1`/`speak.v2` shut down with `{"type":"Close"}` via `sendClose()`, and the helper applies to every streaming resource. Recognize `"Close"` as well, and confirm what close code follows `Close` on the TTS sockets.
- **[S2]** `send()` sets `_terminalMessageSent = true` before checking whether the transport is open, so a `CloseStream` that was queued and never delivered still marks the stream terminal — a genuine network failure afterwards is then treated as an intentional shutdown and the rest of the stream is silently lost. Java #110 tests exactly this ("a rejected CloseStream does not make a no-status close terminal"). Set the flag only on a successful send, or track the transport instance.
- **[S3]** `shouldReconnect` and `off()` are new public API with no README entry. While B1 stands, `shouldReconnect` is not a nicety — it is the only way to close a Flux stream without a stray reconnect, and it is demonstrated only inside three example files.

## Nit

- **[N1]** `src/version.ts` and `package.json` move to 5.11.1 while the release-please manifest stays at 5.11.0. Release-please will compute 5.12.0 from this `feat:` commit, so 5.11.1 never publishes. No action needed — worth a line in the PR body so the next reviewer does not re-flag it.

## Questions

- Does Fern's `websocketHandlerMode` have a `"replace"` value, and if so does it still emit `off()`? If it does, the config knob beats five permanently frozen socket files.
- B1 is a pre-existing gap, not a regression — 1005 reconnected before this regen too. Happy to take this as "merge the regen, fix the default path in a follow-up," provided the PR body and release notes drop the native-terminal-close claim.

GregHolmes:
Addressed the review findings in aa4ef92 and the preceding commits.

- B1: native Listen V2 now tracks a successfully sent CloseStream and treats the following production 1005 close as terminal by default. The live check passes without a caller callback.
- B2: async iteration is registered ahead of the replacing user callback, and message callback errors are logged without preventing iteration delivery.
- S1/S2: custom transports recognize Close as terminal and mark CloseStream only after it has been accepted for sending; a rejected terminal send retains reconnect behavior.
- S3: README documents off() and shouldReconnect.
- N1: PR description now documents the generated 5.11.1 source stamp and the expected 5.12.0 Release Please release.

dg-coreylweathers (changes requested):
Two blocking items before merge. I applied and tested both patches locally against the live API and a clean container build, then reverted so the change is yours — the diffs below are known-good.

Everything else in the regen verified: your fixes for the previous round's two blockers both hold, all 19 manual patches survived, and the new Agent surface lands correctly.

## [B1] The next regen will delete the test guarding this PR's new ws.ts patch

`AGENTS.md` records `src/core/websocket/ws.ts` as newly frozen and names `tests/unit/websocket/ws.test.ts` as its regression coverage. That test is new in this PR and is not in `.fernignore`. Per AGENTS.md's own rule the generator honors only the exact listed paths, so the next regen deletes it and the `ws.ts` patch loses its only guard — silently, showing up as a drop in the test count. The sibling subdirectories (`tests/unit/core`, `tests/wire/websocket`) are all listed. No customer impact today; this is bookkeeping that goes wrong later.

Fix — one line in `.fernignore`, after `tests/unit/websocket-wrappers.test.ts` (line 55), keeping the list sorted:

```
tests/unit/websocket
```

Worth naming it in the `tests/unit/*` bullet in AGENTS.md too, alongside the other suites, so the two files still read as a matched pair.

## [B2] A callback that throws on close or error kills the host application

A `message` callback that throws is caught, logged, and the other handlers still run. A `close` or `error` callback that throws is not: the generated dispatch loop calls each handler unguarded, so the exception aborts the remaining handlers for that event and escapes the loop with nothing above it to catch it. In Node that terminates the process.

**Expected:** a throwing application callback affects only that callback, identically on all four events — the behavior `setupBinaryHandling` already gives `message`.

**Observed:** on `listen.v2` against production, `socket.on("close", () => { throw ... })` printed the error and exited the process. The developer's application dies at the end of a normal transcription session.

This is pre-existing rather than introduced here — the same probe on `main` also throws uncaught, and there it additionally hangs `for await`, which this PR does fix. Raising it now because the five socket files carrying the loop are already unfrozen and hand-patched by this PR, so a follow-up would have to reopen the same five files.

Fix — in all five frozen socket files, wrap the handler call in each of the four dispatch loops (`handleOpen`, `handleMessage`, `handleClose`, `handleError`). Shown for close; same shape for the other three with the event name in the message:

```ts
    private handleClose: (event: core.CloseEvent) => void = (event) => {
        for (const handler of [...this.eventHandlers.close]) {
            try {
                handler(event);
            } catch (error) {
                console.error("Deepgram WebSocket close handler failed", error);
            }
        }
    };
```

One gotcha: do **not** add a `biome-ignore lint/suspicious/noConsole` comment. `noConsole` is not enabled in this repo's `biome.json`, so biome reports the suppression itself as `suppressions/unused` — I hit this and it added 20 warnings. (The three pre-existing ones in `CustomClient.ts` and `ws.ts` are already warnings for the same reason; worth deleting separately.)

Also extend the freeze notes so the next regen re-applies this — the `.fernignore` comment above the four socket entries, and the 5-socket bullet in AGENTS.md.

Regression coverage to add in `tests/unit/websocket-sockets.test.ts`: a `describe` block asserting a throwing `open`/`message`/`close`/`error` handler neither escapes the dispatch loop nor prevents a co-registered handler from running, across four socket classes. Mine came to 8 tests; happy to hand over the exact block if useful.

Worth reporting the unguarded loop upstream to Fern — it is generator output, so every regen reintroduces it.

### How I verified the fix

- Live, production `listen.v2`, throwing close callback: `uncaught=0`, failure logged, `for await` still terminates. Before the patch the same probe exited the process.
- No regression: `tests/manual/flux-terminal-close.ts` and `tests/manual/agent-defer-until-eot.ts` both still pass against production.
- Clean `node:24` container: build, 1,280 unit tests (up 8), 160 wire tests, typecheck and lint all pass, with lint holding at 559 warnings — identical to the pre-change baseline, so no new lint debt.

## What I checked that passed

- **Does a Flux stream stay closed now?** Yes. Production, no caller policy: `sendCloseStream()` → close code 1005 → one `open` event, no reconnect. Your `flux-terminal-close.ts` passes.
- **Is scoping the policy to listen.v2 correct?** Yes. I probed the others: `listen.v1 sendCloseStream` and `speak.v1 sendClose` both close with 1000, so no other resource needs it. The dead `closeStreamSent` you removed from `WrappedListenV1Client` in 3a8bfed was the right call.
- **Is the async-iteration starvation fixed?** Yes for `message` — on `main` that probe crashed the process; on this branch both throws are logged and `for await` finishes.
- **Did the manual patches survive?** All 19, with no `.bak` leftovers and vitest/@vitest/coverage-v8 still identical exact pins.
- **`on(event, undefined)`** still clears the handler — good catch widening the parameter back from the generator's `NonNullable`, that would have been a silent source break.
- **New Agent surface:** `FunctionCallCancelled` is in the response union and exported; production Voice Agent returned `SettingsApplied` for `defer_until_eot`.
- **Version:** 5.11.1 source stamp with the manifest at 5.11.0 is fine — merge commit `96fca66 feat(regen): … (#546)` produced the 5.11.0 minor the same way, so the `feat(regen):` title will cut 5.12.0.

## Non-blocking

- `FunctionCallCancelled` and `defer_until_eot` are documented nowhere outside generated JSDoc — no README, no example. The `Send no FunctionCallResponse for this id` contract is something a developer implementing function calling has to know, and the Python sibling documents the event in its `reference.md`, so this is also a parity gap.
- `CustomClient.ts:1108` `console.error` is unconditional and bypasses the SDK's own `ILogger` (which defaults to silent). Same applies to the 20 new call sites in B2 — a follow-up could route them all through the logger.
- `_send` marks `_terminalMessageSent` before the send promise settles, so a close arriving first still reads as an intentional shutdown (custom transports + 1005 only).
- README's `shouldReconnect` paragraph does not mention that the automatic Flux terminal-close default is skipped when a custom `transportFactory` is configured.

## Still open from the last round

`.fern/metadata.json` now records `"websocketHandlerMode": "accumulate"`, which confirms the generator has a knob for exactly the behavior the five frozen socket files exist to undo. If a `"replace"` value exists and still emits `off()`, that one config line retires five permanently frozen files and the hand-maintained dispatch loops inside them.

GregHolmes:
Addressed the remaining custom-transport terminal-send race in 23e1dbc. A no-status close now waits for the matching terminal send: fulfillment remains terminal, while rejection reconnects normally.\n\nValidation passed locally: make build, make test, make lint (559 existing warnings), and make typecheck-tests. An independent DevRel re-review found no blocking findings, should-fixes, or nits.\n\nGitHub checks are green apart from the still-pending examples job.

dg-coreylweathers (changes requested):
B1 and B2 are fixed — confirmed in the code, not from the summary. `.fernignore:56` carries `tests/unit/websocket`, all four dispatch loops in the five socket files isolate throwing handlers, and README:50 now documents the `FunctionCallCancelled` and `defer_until_eot` contract. Clean `node:24` container: build, 1,287 unit + 160 wire tests, typecheck and lint all pass at the 559-warning baseline.

One new blocking item, introduced by `23e1dbc` — the commit that addressed the non-blocking `_terminalMessageSent` race.

## [B3] Close event races the terminal send, and loses when the send never settles

`23e1dbc` defers the close event while a terminal send is in flight: with a pending `CloseStream`, no caller `shouldReconnect`, and close code 1005, `_handleClose` stores `_pendingTerminalClose` and returns without emitting. The close is emitted only later from `_settleTerminalMessage`. Nothing bounds that wait, and no other path flushes the deferred close — `close()` returns early once `_transport` is already undefined, and `_handleError` clears the pending state without emitting it.

**Expected:** the socket reports its close exactly once, whatever the terminal send does. Close delivery is how the SDK tells the caller the session ended; it should not be gated on an unrelated promise.

**Observed:** with a custom transport whose `send()` promise stays pending, the `close` handler never fires and `for await` never terminates. Reproduced against the built SDK with a fake transport:

| Terminal send | Close delivered |
| --- | --- |
| resolves, then 1005 | yes, `[1005]` |
| rejects, then 1005 | yes |
| **never settles, then 1005** | **no — `closes === []`** |
| **same, then the caller calls `socket.close()`** | **no — still `[]`** |

Narrow scope — it needs a `transportFactory` (SageMaker and similar), a `send()` returning a promise, no caller `shouldReconnect`, and a 1005 close. But the failure is a silent hang with no diagnostic and no escape hatch for the application, which is the same class of defect as B2, reintroduced in the custom-transport path. The two tests added in `23e1dbc` cover the resolve and reject paths; neither covers a send that never settles.

**Recommended fix — the quick one: revert `23e1dbc`.** The race it closes (a queued `CloseStream` that never reached the wire being treated as intentional, dropping the tail of a stream) is milder than the hang it opens, and it was a non-blocking should-fix that can ship separately. That puts the PR back at head `38742af`, where both blockers are resolved and I have nothing further.

If you would rather keep it: stop gating close *delivery* on the send. Emit the close immediately and let the terminal-send outcome influence only the reconnect decision, with a bounded wait (`Promise.race` against a short timer) so a stuck transport degrades to the previous behavior instead of silence. Either way, please add a test for a terminal send that never settles.

## Unchanged from the last round, all non-blocking

- `CustomClient.ts:1108` `console.error` is unconditional and bypasses the SDK's own `ILogger` (which defaults to silent); same for the 20 new call sites in the B2 fix.
- README's `shouldReconnect` paragraph does not mention that the automatic Flux terminal-close default is skipped when a custom `transportFactory` is configured.
- Still open: `.fern/metadata.json` records `"websocketHandlerMode": "accumulate"`. If the generator accepts `"replace"` and still emits `off()`, that one line retires five permanently frozen socket files.

GregHolmes:
Corey - addressed B3 in `9f0e5c0` by reverting `23e1dbc`, per the recommended quick fix. Close events are no longer gated on an unsettled terminal-send promise, so custom transports cannot leave `close` handlers or async iteration hanging.

At `9f0e5c0`, the tree was byte-for-byte identical to the previously reviewed `38742af` tree. Two later commits are also included in this PR:

- `80a0cfa` (`fix: preserve synchronous terminal closes`) records terminal intent before sending, so a custom transport that closes synchronously is not treated as recoverable.
- `f24a206` (`docs: clarify custom transport reconnects`) documents the native/custom reconnect behavior.

Validation on current head `f24a206`: `make build`, `make test` (1,287 unit and 160 wire tests), `make lint` (559 existing warnings), and `make typecheck-tests` passed. All GitHub checks are now green.

dg-coreylweathers (approved):
Approving. Round 4 at f24a206, no blocking findings and no should-fixes.

B1, B2 and B3 are all fixed, confirmed in the code rather than from the summary: `.fernignore:56` carries `tests/unit/websocket`, all five socket files isolate throwing handlers in each of their four dispatch loops, and the B3 revert is byte-exact (`git diff --stat 38742af 9f0e5c0` is empty).

The two commits after your last comment both hold up. `80a0cfa` records terminal intent before `send()` so a synchronously-closing custom transport is no longer misread as recoverable, and it cannot reintroduce B3 — close is emitted unconditionally in `_handleClose`, and `custom-client-internals.test.ts:226` proves close arrives and async iteration terminates when the send promise never settles. Stale state can't cross a reconnect either: both the `catch` and the rejection handler guard on `this._transport === transport`.

All four claims in the new README paragraph check out against the code, and I confirmed the native/custom split is exhaustive with a live probe: `speak.v2` (Flux TTS) and `speak.v1` (Aura) both close with 1000 after `Close`, so Flux Listen V2 really is the only resource that needs the 1005 terminal default.

Gate in a clean node:24 container: build OK, 1,287 unit + 160 wire tests pass, typecheck passes, lint exits 0 at 559 warnings — the same baseline as before this PR. `tests/manual/flux-terminal-close.ts` and `tests/manual/agent-defer-until-eot.ts` both pass against production at this head.

Semver: nothing narrows a type, drops an optional marker, or removes a union member, and the one behavior change (native listen.v2 no longer reconnecting after an intentional CloseStream) removes a reconnect the caller never asked for. `feat(regen)` cutting 5.12.0 is right.

One nit: your latest comment says the tree is byte-for-byte identical to 38742af. That was true at 9f0e5c0, but 80a0cfa and f24a206 landed after it and add 50 lines including a change to `_send` — worth naming them in a comment and in the PR body's Validation list so the next reader doesn't skip them.

Still open for you, as a follow-up rather than a merge condition: `.fern/metadata.json` records `"websocketHandlerMode": "accumulate"`, which is exactly the behavior the five frozen socket files exist to undo. If the generator accepts `"replace"` and still emits `off()`, that one line retires five permanently frozen files and the 20 hand-patched dispatch loops inside them. The unguarded loop is also generator output, so it's worth reporting upstream to Fern.

Unchanged non-blockers: the 21 `console.error` sites bypass the SDK's own ILogger (silent by default), and the pre-settle terminal-send race is now a deliberate, tested choice — the right call, given the alternative was B3's silent hang.

PR #558, opened 2026-09-18
fix(skills): use folded block scalar so SKILL.md frontmatter parses

dg-coreylweathers:
## Problem

`.agents/skills/deepgram-js-text-intelligence/SKILL.md` has YAML frontmatter that **does not parse**. The `skills` CLI therefore skips the skill, and this repo installs **6 of its 7 skills**.

The `description:` value was an unquoted plain scalar containing a colon-space sequence:

```
description: Use when writing or reviewing JavaScript/TypeScript ... with `body: { text }` or `body: { url }`. ...
```

In YAML, `body: { text }` inside a plain scalar reads as the start of a nested mapping, so the parse aborts. Exact error from `yaml@2.8.2` (the parser the `skills` CLI itself uses):

```
Nested mappings are not allowed in compact mappings at line 2, column 14:

description: Use when writing or reviewing JavaScript/TypeScript in this repo t…
             ^
```

The offending substrings are `` `body: { text }` `` and `` `body: { url }` ``.

## Why this went unnoticed

The installer treats an unparseable skill as **non-fatal**: it emits one dim `⚠ Skipped` warning, drops the skill, reports a lower count, and **still exits 0**. Nothing fails, so the only visible symptom is a skill count that is quietly one short — and in CI, where only the exit code is checked, there is no symptom at all.

Note also that `deepgram/skills`'s `.claude-plugin/marketplace.json` lists **all 7** skills for this repo, including `.agents/skills/deepgram-js-text-intelligence`. So the published manifest and installable reality currently disagree.

## Proof (before / after)

Run in a clean `node:22-alpine` container, same command both times:

**Before** (`main`):

```
◇  Source: /repo
⚠ Skipped /repo/.agents/skills/deepgram-js-text-intelligence/SKILL.md — YAML parse error: Nested mappings are not allowed in compact mappings at line 2, column 14:
◇  Found 6 skills
EXIT=0
```

**After** (this branch):

```
◇  Source: /repo
◇  Found 7 skills
EXIT=0
```

Confirmed against the published repo too — `npx -y skills@latest add deepgram/deepgram-js-sdk --list` reports `Found 6 skills` and exits 0, and a real `--yes` install writes only 6 skill directories.

## Fix

Convert `description` to a `>` folded block scalar, matching the house style already used by `deepgram/skills`'s capability skills (`skills/speech-to-text/SKILL.md` et al.), which is why those survive colons.

**The description text is unchanged.** Verified by re-parsing and comparing to the original string — identical apart from the single trailing newline a `>` folded scalar adds. No rewording or shortening, since these descriptions drive skill matching.

## Verification

A validator was run over the frontmatter of every `.agents/skills/*/SKILL.md` in all six public Deepgram SDK repos (42 skills) using `yaml@2.8.2`:

- Before: 2 failures repo-wide — this file and `deepgram-rust-sdk`'s `deepgram-rust-voice-agent` (fixed in deepgram/deepgram-rust-sdk#182).
- After: **42/42 parse**, and every `name` still matches its directory name.

`deepgram-python-sdk`, `deepgram-java-sdk`, `deepgram-go-sdk` and `deepgram-dotnet-sdk` are all clean — no action needed there.

Leaving this for the JS SDK owners to review and merge.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


GregHolmes:
One optional, non-blocking nit: `description: >` adds a trailing newline to the parsed frontmatter value. The current change fixes installation correctly, but `description: >-` would preserve the original description string exactly.

Verified locally with `yaml@2.8.2`: all seven `.agents/skills/*/SKILL.md` files parse and their `name` values match the directory names.

GregHolmes (approved):

